PR #37, opened 2023-07-20
perf(nebula): optimize AT128 decoder performance

mojomex:
## PR Type

- Improvement

## Related Links

--

## Description

This PR improves performance of the Hesai AT128 decoder by approx. 23% without changing the logical output.
To achieve this, three changes have been made:
* Decoding points directly to `scan_pc_` and `overflow_pc_` instead of temporary buffers and re-using the buffers (~ 11%point improvement)
* Removing unused copy operation for incoming scan packets (~ 9%point improvement)
* Combining azimuth and elevation sin/cos lookup tables as they were exact duplicates (~ 4%point improvement)

Points were decoded in `convert(block_id)` to a temporary PointCloud, the contents of which were then copied to `scan_pc_` or `overflow_pc_` respectively. This caused avoidable latency. Now, depending on whether the current block being decoded is in a new frame/scan or not, `convert(block_id, pcl)` decodes those points directly into the right buffer.

Further, `overflow_pc_` was reallocated on each full scan, causing avoidable heap allocations.
Because each incoming `PandarScan` message always contains a full scan (albeit not necessarily aligned to our phase setting), `scan_decoder->get_pointcloud()` will be called exactly once in `ConvertScanToPointCloud`. Thus, it is guaranteed that one of `scan_pc_` and `overflow_pc_` can always be used for decoding, and the other as the output buffer.
Once a scan is complete, the buffers are swapped and the `overflow_pc_` is `clear`ed (not reallocated).

No changes have been made to the base class, the new function is an overloaded version of the existing one (which has been changed to throw a "not implemented" error).

f7249549-193f-4193-a43a-ef47260f2416

|                     |Before|Optimized|
|-----------------|---------|--------------|
|duration AVG        |7.972725 ms      |6.147384 ms|
|duration STD         |0.759695       |0.613904|
|duration AVG % rel. to before        |100.000000      | **77.105188** |
|duration STD % rel. to before    |100.000000      |80.809372|

All measurements have been made on an Intel Core i7-12700H with
* TurboBoost disabled
* All cores set to 2.3 GHz
* Thread pair 2/3 isolated
* Nebula `taskset` to core 2

Measurements were repeated 3 times and ran for 37s each (~370 iterations).

## Review Procedure

To confirm logical equality with the previous implementation, a rosbag of recorded AT128 packets can be used.
To confirm performance improvement, I provide the scripts and logging that yielded the above charts:
```
scripts/
    test_runner.bash  # runs the driver for 3 iterations on a given rosbag
    plot_times.py     # takes the logs from test_runner.bash and creates a plot like the one above
```
The test runner locks/unlocks core frequencies and `taskset`s nebula onto the specified core(s). Ideally, the taskset cores are isolated (cf. [Autoware performance measurement guide](https://autowarefoundation.github.io/autoware-documentation/main/how-to-guides/integrating-autoware/tuning-parameters-and-performance/evaluating-real-time-performance/#prepare-separated-cores))

*PR Author should check the checkboxes below when creating the PR.*
- [x] Assign PR to reviewer

*Reviewers should check the checkboxes below before approval.*
- [x] (Optional) Unit tests have been written for new behavior

*PR Author should check the checkboxes below before merging.*
- [x] All open points are addressed and tracked via issues or tickets

## CI Checks

- *Build and test for PR*: Required to pass before the merge.

### DIFF
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
@@ -31,6 +31,11 @@ const uint16_t MAX_AZIMUTH_DEGREE_NUM = 36000;
 const uint16_t LIDAR_AZIMUTH_UNIT = 256;
 const uint32_t MAX_AZI_LEN = 36000 * 256;
 
+/// @brief The resolution of one scan (in single return mode)
+const uint16_t SCAN_POINTS_NUM = 1200 * LASER_COUNT;
+/// @brief This sensor supports a maximum of 2 returns (= dual return)
+const uint16_t MAX_RETURN_COUNT = 2;
+
 /// @brief Hesai LiDAR decoder (AT128)
 class PandarATDecoder : public HesaiScanDecoder
 {
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
@@ -79,6 +79,10 @@ private:
   /// @param block_id target block
   /// @return Point cloud
   drivers::NebulaPointCloudPtr convert(size_t block_id) override;
+  /// @brief Convert to point cloud (without temporary buffer)
+  /// @param block_id target block
+  /// @param out_pc Point cloud to append the decoded points to
+  void convert(size_t block_id, NebulaPointCloudPtr & out_pc);
   /// @brief Convert to point cloud for dual return
   /// @param block_id target block
   /// @return Point cloud
@@ -93,13 +97,8 @@ private:
   std::array<float, BLOCKS_PER_PACKET> block_offset_dual_{};
   std::array<float, BLOCKS_PER_PACKET> block_offset_triple_{};
 
-  std::vector<float> m_elevation_rad_map_;
-  std::vector<float> m_sin_elevation_map_;
-  std::vector<float> m_cos_elevation_map_;
-
-  std::vector<float> m_azimuth_rad_map_;
-  std::vector<float> m_sin_azimuth_map_;
-  std::vector<float> m_cos_azimuth_map_;
+  std::vector<float> m_sin_map_;
+  std::vector<float> m_cos_map_;
 
   Packet packet_{};
 
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
@@ -32,7 +32,7 @@ const uint16_t LIDAR_AZIMUTH_UNIT = 256;
 const uint32_t MAX_AZI_LEN = 36000 * 256;
 
 /// @brief The resolution of one scan (in single return mode)
-const uint16_t SCAN_POINTS_NUM = 1200 * LASER_COUNT;
+const uint32_t SCAN_POINTS_NUM = 1200 * LASER_COUNT;
 /// @brief This sensor supports a maximum of 2 returns (= dual return)
 const uint16_t MAX_RETURN_COUNT = 2;
 
--- a/nebula_decoders/src/nebula_decoders_hesai/decoders/pandar_at_decoder.cpp
+++ b/nebula_decoders/src/nebula_decoders_hesai/decoders/pandar_at_decoder.cpp
@@ -32,19 +32,12 @@ PandarATDecoder::PandarATDecoder(
     azimuth_offset_[laser] = calibration_configuration->azimuth_offset_map[laser];
   }
 
-  m_elevation_rad_map_.resize(MAX_AZI_LEN);
-  m_sin_elevation_map_.resize(MAX_AZI_LEN);
-  m_cos_elevation_map_.resize(MAX_AZI_LEN);
-  m_azimuth_rad_map_.resize(MAX_AZI_LEN);
-  m_sin_azimuth_map_.resize(MAX_AZI_LEN);
-  m_cos_azimuth_map_.resize(MAX_AZI_LEN);
+  m_sin_map_.resize(MAX_AZI_LEN);
+  m_cos_map_.resize(MAX_AZI_LEN);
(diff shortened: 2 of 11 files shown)


scripts/plot_times.py
mojomex, line 1: (For validation purposes only) will be removed once accepted.

scripts/test_runner.bash
mojomex, line 1: (For validation purposes only) will be removed once accepted.

nebula_ros/src/hesai/hesai_decoder_ros_wrapper.cpp
mojomex, line 1: (For validation purposes only) changes will be reverted once accepted.

nebula_ros/include/nebula_ros/hesai/hesai_decoder_ros_wrapper.hpp
mojomex, line 1: (For validation purposes only) changes will be reverted once accepted.

scripts/plot_times.py
amc-nu: Instead of deleting the file, can you please:
1. Parametrize this script?
2. Write in the README how to use it?

nebula_ros/src/hesai/hesai_decoder_ros_wrapper.cpp
```diff
@@ -98,6 +97,9 @@ void HesaiDriverRosWrapper::ReceiveScanMsgCallback(
       rclcpp::Time(SecondsToChronoNanoSeconds(std::get<1>(pointcloud_ts)).count());
     PublishCloud(std::move(ros_pc_msg_ptr), aw_points_ex_pub_);
   }
+
+  auto runtime = std::chrono::high_resolution_clock::now() - t_start;
+  RCLCPP_INFO(get_logger(), "AAA {'d_total': %lu, 'n_out': %lu} BBB", runtime.count(), pointcloud->size());
```
amc-nu, line 102: Can you please convert this to a DEBUG_INFO?

nebula_decoders/src/nebula_decoders_hesai/decoders/pandar_at_decoder.cpp
```diff
@@ -58,13 +51,15 @@ PandarATDecoder::PandarATDecoder(
 
   scan_pc_.reset(new NebulaPointCloud);
   overflow_pc_.reset(new NebulaPointCloud);
+  scan_pc_->reserve(1200 * 128);
```
amc-nu, line 54: Can you please convert these numbers to `constexpr` in the header and assign them a name?

amc-nu (changes requested):

nebula_decoders/src/nebula_decoders_hesai/decoders/pandar_at_decoder.cpp
mojomex: Maybe it is better to get rid of the reserve altogether:
The 1200*128 case is only valid for single return and would be double for dual return.
Thus, we could either:
* Reserve space for dual return (wasteful if only single return is used)
* Reserve space for single return (vector would grow on first dual return scan)
* Reserve nothing (vector grows on first scan)

I am in favor of the last option as it is the most concise and does not decrease performance except for the first scan.

nebula_ros/src/hesai/hesai_decoder_ros_wrapper.cpp
mojomex: Done. This necessitated a `debug_logging` parameter for the `nebula_launch.py` file to actually enable debug output.

scripts/plot_times.py
mojomex: It is parameterized/documented now :+1:

nebula_decoders/src/nebula_decoders_hesai/decoders/pandar_at_decoder.cpp
amc-nu: > Is Nebula supporting the switching of return modes on the fly?

It should be supported inherently but not tested.
I would be in favor of option one as well.

scripts/profiling_runner.bash
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/bash
```
amc-nu, line 1: Can you please add how to use this and the plot scripts at the end of the README?

nebula_decoders/src/nebula_decoders_hesai/decoders/pandar_at_decoder.cpp
mojomex: Done :+1: 
There are now constants for scan resolution (width * LASER_NUM) and MAX_RETURNS which are then used to reserve space for the maximum number of points that can occur.

scripts/profiling_runner.bash
mojomex: Moved it to the end :heavy_check_mark:

nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
```diff
@@ -31,6 +31,11 @@ const uint16_t MAX_AZIMUTH_DEGREE_NUM = 36000;
 const uint16_t LIDAR_AZIMUTH_UNIT = 256;
 const uint32_t MAX_AZI_LEN = 36000 * 256;
 
+/// @brief The resolution of one scan (in single return mode)
+const uint16_t SCAN_POINTS_NUM = 1200 * LASER_COUNT;
```
amc-nu, line 35: Upgrade to 32-bit; this variable overflows.

amc-nu:
Other than my comment above, the PR works great!
Thank you. @mojomex 
cc @drwnz

nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/pandar_at_decoder.hpp
mojomex: Thanks, overlooked that one!

amc-nu (approved):

amc-nu:
@mojomex can you please attend to the spelling errors on the [spell-check job](https://github.com/tier4/nebula/actions/runs/5785120036/job/15696013740?pr=37)?

mojomex:
Most of the warnings were the names of Bash commands, added them to the dictionary. `stds` for standard deviations is also commonly used in Python/Pandas, added to the dictionary as well.

amc-nu (approved):

PR #39, opened 2023-07-24
feat(sync_angle): hesai. Enable Clock Sync Features

amc-nu:
## PR Type

- New Feature

## Related Links

https://tier4.atlassian.net/browse/RT1-3110?atlOrigin=eyJpIjoiZjIyZDQ1ODA1NmZmNGQyNWIzOTNmZTg5YzlmMmVhY2MiLCJwIjoiaiJ9

## Description

- Sets lidar sync angle on launch on Hesai drivers.
- Sets PTP as the default clock source
- Sets PTP config to IEEE1588v2 over UDP/IP as default.

## Review Procedure

- Set a value in the `scan_phase` parameter in the launch file. 
- Launch the driver.
- Verify the value has changed using Lidar's web UI.

## Remarks

There is a known bug on the Pandar40P where the Lidar only responds to the first TCP command causing timeout on the subsequent commands. The only known way to solve this is to power cycle the Lidar.

~NOTE: This won't work with the AT128 since it doesn't support PTP setup through TCP commands.~
Added support for the AT128 via HTTP.


drwnz (commented):
Tested on AT128 and Pandar40P. As noted, Pandar40P needs a device restart before settings are applied.
Sync angle settings are working as required.

drwnz (approved):

PR #42, opened 2023-07-27
refactor(nebula_decoders_hesai): implement generic Hesai decoder

### LINKED ISSUE #38: Generalize decoder implementation
### Description

Currently, similar models from the same vendor share very little of their decoder code despite their decoders working largely the same. This issue calls for a generalized decoder (at least per vendor) that is only specialized where necessary.

### Purpose

* De-duplicated code is much more maintainable/understandable
* Optimizations/bugfixes for one lidar decoder would automatically apply to all decoders

### Details

For Hesai lidar packets:
* The pre-header and header fields are the same format (except for Pandar-64 (short header) and Pandar-40P (no header))
* The block (≈ lidar column) format is identical except for an optional fourth (mostly unused) byte
* The body (group of blocks) format is the same except for number of blocks, fine azimuth and checksum fields
* Formats in the tail are wildly different. However, only a few fields are used by Nebula and those do have the same format
* Return modes can be handled in a generalized way (format is consistent across models)

The overall algorithm of decoding the packets is basically identical (parameterization is of course model-dependent).

### Possible approaches

Implement the decoding algorithm fully generically, using specific model definitions (as they are already largely present) for parameterization. The model definitions should re-use structs for headers/blocks whereever possible. If the structs are packed, the `memcpy`-based parser from [`pandar_128_e4x_decoder`](https://github.com/tier4/nebula/blob/main/nebula_decoders/src/nebula_decoders_hesai/decoders/pandar_128_e4x_decoder.cpp) can be used for all models. Accessor functions to the specific struct locations can then provide the necessary abstraction.

### LINKED ISSUE #48: Hesai unit test pointclouds contain duplicate points
## Description

The reference `.pcl` files in the Hesai unit tests contain large numbers of identical points, often close to half of the total points in the scan:

| Sensor model | Total # points | duplicate # points | % duplicate |
|-|-:|-:|-:|
| 40P | 108663 | 51941 | 47.8 |
| 64 | 179121 | 85501 | 47.7 |
| AT128 | 10205 | 0 | 0.0 |
| QT64 | 64452 | 30743 | 47.7 |
| XT32 | 121648 | 58627 | 48.2 |
| XT32M | 13404 | 0 | 0.0 |

Unit tests are passing only because the `dual_return_distance_threshold` parameter is set to `0.0`. This is also the case in `nebula_launch.py` but is set to a default value of `0.1` in `hesai_launch_all_hw.xml`.

## Expected behavior

There are no duplicate points in the reference data of unit tests, and decoders should be tested (**and launched!**) with a `dual_return_distance_threshold > 0` or identical point checks (such as AT128 has).

Reasoning: Duplicated points provide no value but take up memory, thus increasing DDS latency, processing resources etc.

## Actual behavior

There are duplicate points in the reference data, and while decoders have code to filter by `dual_return_distance_threshold`, this parameter is never set in the current `nebula_launch.py`, thus implicitly being initialized to `0` and making the filter code do nothing. In the unit tests, it is explicitly set to `0`.

## Steps to reproduce

1. In all `hesai_ros_decoder_test_<model>.cpp` unit test files, replace the `void checkPCDs(nebula::drivers::NebulaPointCloudPtr, pcl::PointCloud<pcl::PointXYZ>::Ptr)` function by the following:
```c++
void checkPCDs(nebula::drivers::NebulaPointCloudPtr pc, pcl::PointCloud<pcl::PointXYZ>::Ptr ref_pc)
{
  /*
   * Check for duplicate points in reference pointcloud
   */

  size_t n_neighbor_duplicates = 0;
  for (size_t i = 1; i < ref_pc->size(); ++i) {
    auto & p1 = ref_pc->points[i - 1];
    auto & p2 = ref_pc->points[i];
    if (p1.x == p2.x && p1.y == p2.y && p1.z == p2.z) {
      n_neighbor_duplicates++;
    }
  }

  std::cout << "ref_pcl n_duplicates: " << n_neighbor_duplicates << std::endl;
  std::cout << "ref_pcl n_total:      " << ref_pc->points.size() << std::endl;

  /*
   * Do the actual unit tests
   */

  EXPECT_EQ(pc->points.size(), ref_pc->points.size());
  for (uint32_t i = 0; i < pc->points.size(); i++) {
    auto p1 = pc->points[i];
    auto p2 = ref_pc->points[i];

    EXPECT_FLOAT_EQ(p1.x, p2.x);
    EXPECT_FLOAT_EQ(p1.y, p2.y);
    EXPECT_FLOAT_EQ(p1.z, p2.z);
  }
}
```

2. Build and run the unit tests
3. Observe the output:
```shell
ref_pcl n_duplicates: 58627
ref_pcl n_total:      121648
```

### LINKED ISSUE #49: `MIN_RANGE`, `MAX_RANGE` and `dual_return_distance_threshold` handled differently across sensors
## Description

The checks whether a point is below/above a minimum/maximum range and the check whether two points are within a threshold of each other in dual return are different across sensors:

|Sensor Model | Single-Return |  | Dual-Return |  | |
|-|-|-|-|-|-|
| | Min-Distance | Max-Distance | Min-Distance | Max-Distance | Dual-Return Threshold |
|QT64 | **>** | <= | **>** | <= | >=|
|QT128 | >= | <= | >= | <= | **>**|
|64 | **>** | <= | >= | <= | >=|
|40P | **--** | **--** | >= | <= | >=|
|AT128 | >= | <= | >= | <= | **--**|
|XT32 | >= | <= | **>** | <= | **--**|
|XT32M | >= | <= | >= | <= | **--**|
|128 | >= | <= | >= | <= | **>**|

As can be seen, the majority of comparisons uses `>=` or `<=` while some exceptions use `>` or `<` and some sensors use no distance/threshold check at all (`--`).

This makes a consistent yet readable implementation of #42 practically impossible (an extra function override per exception would be required). A unified implementation fails unit tests as there are actually some points exactly on (`=`) those boundaries.

I propose to use the `>=` and `<=` operators for all checks since they are used by the majority of sensors, and because `MIN_RANGE` is interpreted most intuitively with "the minimum range to keep point P", etc.

## Expected behavior

All sensors use the same operators for the same checks, all sensors employ the same checks (!).

## Actual behavior

Some sensors use a non-`=` operator and some omit distance/threshold checks completely.

mojomex:
Implement a generic Hesai decoder that can handle all Hesai sensor models.
Fixes #48, fixes #49 and closes #38.

## PR Type

- Improvement
- Refactoring

## Related Links

#38 -- the issue for this PR
#48 -- the first issue that made the unit tests of this PR fail
#49 -- the second issue that made the unit tests of this PR fail

## Description

Since sensors from the same vendor often follow similar conventions when it comes to packet structure and data processing steps, a generic decoder can be used for most of the decoding work.
This PR propses such a generic decoder for all Hesai sensors currently supported.

#### Packet formats

For all handled Hesai sensors, the packet structure follows this rough format:
1. (optional) header: static sensor info and feature flags
2. body: point data
3. tail and other appendices: timestamp, operation mode info

#### Decoding steps

For all handled Hesai sensors, decoding a packet follows these steps:
1. parse (& validate) packet
2. decode and append points to pointcloud, iterating over groups of blocks (corresponding to the same firing for multi-return) or single blocks (single-return)
    1. check if block (group) completes scan and swap output buffers
    2. iterate over units in block and decode
        1. get distance from packet
        2. filter by distance
        3. get and correct azimuth and elevation from packet __*__
        4. compute x, y, z using trigonometry (preferably lookup tables)
        5. compute and correct time offset of point __*__
        6. assign return type to point (handling possible duplicates from multi-return) __*__
        7. append point to pointcloud

The steps marked with __*__ are model-specific:
* angle correction
* timing correction
* return type assignment

#### Angle correction

There are two approaches between all the supported sensors:
* Calibration file based
* Correction file based (currently only used by AT128)

For both approaches, the same sin/cos lookup tables can be computed and used.
However, the resolution and calculation of these tables is different.

##### Calibration based

For each laser channel, a fixed elevation angle and azimuth angle offset are defined in the calibration file.
sin/cos lookup tables are computed at 0.001 deg resolution (the resolution observed in the calibration files) for the whole 360 deg range .

This yields a (360 deg / 0.001 deg) * sizeof(float) = 360000 * 4B = 1.4MB memory footprint per lookup table (of which there are two: sin, cos).

##### Correction based

While azimuth and elevation correction also have a per-channel component, an additional component depending on azimuth AND channel is present.

This leads to more individual azimuth/elevation values than the calibration-based approach, and for the case of the AT128, the lookup table resolution is around 0.00004 deg.

This yields a memory footprint of 36.9 MB per lookup table.

#### Timing correction

Timing correction for all sensors follows the same underlying formula:
Given a scan start timestamp $T_{s_i}$, packet start timestamp $T_{p_j}$, block offset $o_{b_k}$ within the packet and channel offset $o_{c_l}$ within the block, the point identified by $(i, j, k, l)$ has the relative scan timestamp $t_{i,j,k,l} = T_{p_j} - T_{s_i} + o_{b_k} + o_{c_l}$.

The block offset follows a formula linear in the block index for all sensor models which addidionally depends on the number of returns of the currently active `return_mode`.

The channel offset is given as a formula, table or set of tables for all sensors. A few sensors' formula is influenced by factors such as high resolution mode (128E3X, 128E4X), alternate firing sequences (QT128) and near/farfield firing (128E3X).

#### Return types

While there is a wide range of different supported return modes (e.g. single (first), single (strongest), dual (first, last), etc.) their handling is largely the same.
Differences only arise in multi-return (dual or triple) in the output order of the returns, and in the handling of some returns being duplicates (e.g. in dual(first, strongest), the first return coincides with the strongest one).

Here is an exhaustive list of differences:
* For Dual (First, Last) `0x3B`, 128E3X, 128E4X and XT32 reverse the output order (Last, First)
* For Dual (Last, Strongest) `0x39`, all sensors except XT32M place the second strongest return in the even block if last == strongest
* For Dual (First, Strongest) `0x3c`, the same as for `0x39` holds.

For all other return modes, duplicate points are output if the two returns coincide.

#### `HesaiPacket`

Packets are defined as **packed** structs for effortless parsing.
The sensor-specific layout for sensor XYZ is defined in `PacketXYZ` and usually employs an own `TailXYZ` struct.
The header formats are largely shared between sensors.
The packet body (i.e. point data) is mainly parameterized by bytes per point, points per block, and blocks per body. Thus, parameterized generic structs are used. A few skews such as fine azimuth blocks and blocks with a start-of-block (SOB) header exist and are implemented as their own structs.

HesaiPacket diagram

#### `HesaiSensor`

Each sensor model has its own class `PandarXYZ : HesaiSensor<...>` that defines packet type and timing, and return mode handling logic. Angle correction is the same for 90% of sensors and thus outsourced into `AngleCorrector` and subclasses. These are template arguments for `HesaiSensor`.
Return mode handling has a default implementation that is supplemented by additional logic only in 3 sensors.

#### `AngleCorrector`

The angle corrector has three main tasks:
* compute corrected azimuth/elevation for given azimuth and channel
* implement `hasScanCompleted()` logic that decides where one scan ends and the next starts
* compute and provide lookup tables for sin/cos/etc.

The two angle correction types are calibration-based and correction-based. In both approaches, a file from the sensor is used to extract the angle correction for each azimuth/channel.
For all approaches, cos/sin lookup tables in the appropriate size (360deg) and resolution (either 1/1000th deg or 1/25600th deg) are generated.
These resolutions come from the resolutions of the calibration file angle offsets (1/1000th deg) and the fine azimuth resolution of AT128 (1/15600th deg).

#### `HesaiDecoder<SensorT>`

The decoder is in charge of the control flow and shared decoding steps of all sensors.
It is a template class taking a sensor type `SensorT` from which packet type, angle correction etc. are deducted at compile time.
Thus, this unified decoder is an almost zero-cost abstraction.

Its tasks are:
* parsing an incoming packet
* managing decode/output point buffers
* converting all points in the packet using the sensor-specific functions of `SensorT` where necessary

## Remarks

Current progress

| Task | Done? | Notes |
|-|-|-|
| Visually correct output | :heavy_check_mark:  |  |
| Angle correction working | :heavy_check_mark: | currently with slight deviations from the previous output, see comment below |
| Timing correction working | :heavy_check_mark: | incl. hires-mode support for 128E3X/128E4X |
| Return type logic | :heavy_check_mark: | incl. first/strongest becoming first_strongest/secondstrongest for some sensors |
| Documentation | :heavy_check_mark: | API sufficiently documented (imho), design document integrated in nebula docs |
| Finalize design | :yellow_circle: | discussion ongoing |
| Test on realworld data | :heavy_check_mark: | validated 40P, 64, AT128, QT64, XT32 (others not available) |
| Unit tests passing | :heavy_check_mark: | Adapted the original decoders to hotfix #48 and #49 and regenerated unit test reference data. With the new data, tests pass. For details, see comment below |
| Performance measurement | :heavy_check_mark: | Faster or equal to before for all sensors |

- [x] Assign PR to reviewer

## CI Checks


mojomex:
@amc-nu @drwnz The implementation is feature complete and mostly documented, feel free to have a look. Design needs small cleanups in some spots but it should be fairly readable.

mojomex:
Since all unit tests are passing now (*with some caveats, see below), I will mark this PR as ready and open the discussion!

### Unit tests
I modified the original decoders in the following ways and then re-generated the unit test pointclouds using them:
* Use the same comparison operators for min/max range check, dual return threshold in all decoders
* Set `dual_return_distance_threshold = 0.1` in the unit tests (was unset (=0) before) to match the `hesai_launch_all_hw.xml`  launch file
* Change output order for XT32M for multi-return to match the other sensors (was block-major, now is column-major within the return group, i.e. the two returns from one channel will be output next to each other now, whereas they had an offset of #channels before)
* Remove weird logic for checking if a scan is complete in XT32M (involved timing for some reason)
* Fix bug in XT32 where in dual return, points are always checked against the 1st block (instead of the 1/3/5th block for the 2/4/6th block respectively)

Further, I do not test for float equivalence between reference and decoder pointcloud, but allow for some error. Further, I do not test for x/y/z equivalence but rather compute distance, azimuth and elevaion, mainly for easier troubleshooting. This can be changed back once testing is complete.
The permitted deviations from the reference data are currently:
| Variable | Absolute error |
|-|-|
| distance | 1 cm |
| azimuth | 0.001 deg |
| elevation | 0.001 deg |

### Angle correction differences
There is currently a deviation between the angles my code computes and those in the unit test reference data.
I am very sure this comes down to rounding errors either in the original or in my implementation but I am still investigating.

image

amc-nu (approved):

drwnz (approved):
Awesome work!

PR #47, opened 2023-08-08
fix: Update velodyne point stamp calculation

amc-nu:
## PR Type
- Bug Fix

## Related Links

[](https://star4.slack.com/archives/C057JDFS5M2/p1691462406694639?thread_ts=1691455050.384009&cid=C057JDFS5M2)

## Description

Fixes Autoware points point independent timestamp calculation.


drwnz (approved):
Tested working, solves issues with distortion corrector node in Autoware.

PR #56, opened 2023-08-21
fix(nebula_decoders_hesai) unify pointcloud order and filtering of all sensors

### LINKED ISSUE #48: Hesai unit test pointclouds contain duplicate points

### LINKED ISSUE #49: `MIN_RANGE`, `MAX_RANGE` and `dual_return_distance_threshold` handled differently across sensors

mojomex:
Unified output and filtering behavior for all Hesai decoders, fixes #48 (duplicate points in output), fixes #49 (incoherent distance checks).

## PR Type

- Fix
- Improvement

## Related Links

#48 -- duplicate points being output
#49 -- range checks not being consistent

## Description

This PR unifies the behavior of all Hesai decoders to produce outputs with
* no duplicate points 
* the same ordering `return_group > channel > block`
* the same filtering rules
   - `point.distance != 0`
   - `point.distance in [min_range, max_range]`
   - `dist(point1, point2) >= dual_return_distance_threshold` for dual return
* the same precision regarding point distance

The unit tests are regenerated to match the new conventions.

### Detailed changes

Below are the detailed changes made, grouped by sensor.
For details on which comparison operators have been changed, please refer to the table in #49.
For details on how this PR affects output point counts (in the unit test data), see the table in #48.

#### XT32M

* Point output order is changed to `return_group > channel > block` from `block > channel` to match the other sensors
* The check whether a scan is complete is adapted to match those of the other sensors (dropped support for time-based scan slicing)
* Dual return distance threshold checks are added

#### 40P

* Channels are now processed in ascending index order instead of chronological firing order to match other sensors
* The `LASER_RETURN_TO_DISTANCE_RATE` is changed to `double` to match the precision of the calculations of other sensors
* `min_range` and `max_range` distance checks are added

#### XT32

* All dual return types are now correctly interpreted as dual-return (previously only Last+Strongest was recognized)
* Dual return distance threshold checks are added

#### AT128

* All dual return types are now correctly interpreted as dual-return (previously only Last+Strongest was recognized)
* Dual return distance threshold checks are added

*PR Author should check the checkboxes below when creating the PR.*
- [ ] Assign PR to reviewer

*Reviewers should check the checkboxes below before approval.*

*PR Author should check the checkboxes below before merging.*

## CI Checks

- *Build and test for PR*: Required to pass before the merge.


mojomex:
This PR will be made obsolete by #42 but it still serves as documentation on how unit test reference data was changed and how the new, generic decoder handles range checks etc.

amc-nu:
Superseeded by #38

Issue #60, opened 2023-08-29
Enhancement: Add Ring-Based Filter for Removing Reflected and Vehicle Points


mebasoglu:
## Overview
The proposed enhancement introduces a ring-based filtering mechanism to the existing functionality of the repository. This feature aims to remove unwanted points from specified rings within defined angle ranges, specifically targeting reflected and vehicle points directly within the driver. The intention is to elude crop box filters to decrease the delay of the sensing pipeline.

## Details
The key aspects of this feature include:

1. **Point Removal:** The feature allows for the removal of points based on their position within specified rings.
  
2. **Ring and Angle Specification:** Users will be able to define which rings they want to apply filtering to. Each ring will be associated with one or more angle ranges for removal.
  
3. **Angle Ranges:** For each specified ring, users can define multiple angle ranges within which points should be removed. These angle ranges are determined by the starting and ending angles.
  
4. **Fine-Tuned Filtering:** The ability to define multiple angle ranges for a single ring provides fine-tuned control over the filtering process, enabling users to selectively remove points.
  

**Usage Example:**

```xml
<arg name="invalid_point_remove" default="true"/> <arg name="invalid_rings" default="[3, 5, 7, 7]"/> <arg name="invalid_angles_start" default="[120, 250, 180, 230]"/> <arg name="invalid_angles_end" default="[200, 300, 220, 245]"/>
```

**Expected Behavior:** Upon enabling the `invalid_point_remove` option and specifying the relevant `invalid_rings`, `invalid_angles_start`, and `invalid_angles_end` configurations, the ring-based filter will remove points located within the defined angle ranges for the specified rings.

Could you please provide any feedback, suggestions, and guidance on the proposed enhancement @drwnz @amc-nu ?

mebasoglu:
Yes, I thought a decoder level implementation for each sensor, so that we can deactiavte crop box filters to remove vehicle points. 
It could be also useful to remove reflected points from the vehicle's glasses:

![arrowed](https://github.com/tier4/nebula/assets/13589149/591e4c72-d3a3-4761-a1bd-204db0334a57)

mojomex:
This concept has been superseded by the more general #250 mask-based filter. We should be able to make it available for Velodyne pretty quickly, c.f. #251.

PR #69, opened 2023-09-06
Memin/dev/ring based filter

mebasoglu:
## PR Type

- New Feature

## Related Links

The issue: https://github.com/tier4/nebula/issues/60

## Description
This is an implementation for the ring based filter.

## Review Procedure

It is implemented for every decoder part for Velodyne lidars. I created a function `check_invalid_point` in `velodyne_scan_decoder.hpp` and used that function in decoders.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_velodyne/decoders/velodyne_scan_decoder.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_velodyne/decoders/velodyne_scan_decoder.hpp
@@ -192,14 +192,12 @@ public:
   /// @brief Checks if the point is invalid.
   /// @param point The point to check.
   /// @return True if the point is invalid, false otherwise.
-  bool check_invalid_point(const int& channel, const uint16_t& azimuth)
+  bool check_invalid_point(const int & channel, const uint16_t & azimuth)
   {
     if (!sensor_configuration_->invalid_point_remove) return false;
-    for (size_t i = 0; i < sensor_configuration_->invalid_rings.size(); ++i) {
-      if (
-        sensor_configuration_->invalid_rings.at(i) == channel &&
-        azimuth >= sensor_configuration_->invalid_angles_start.at(i) &&
-        azimuth <= sensor_configuration_->invalid_angles_end.at(i)) {
+
+    for (const InvalidRegion & region : sensor_configuration_->invalid_regions) {
+      if (region.ring == channel && azimuth >= region.start && azimuth <= region.end) {
         return true;
       }
     }
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_velodyne/decoders/velodyne_scan_decoder.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_velodyne/decoders/velodyne_scan_decoder.hpp
@@ -188,6 +188,23 @@ public:
   virtual void reset_pointcloud(size_t n_pts) = 0;
   /// @brief Resetting overflowed point cloud buffer
   virtual void reset_overflow() = 0;
+
+  /// @brief Checks if the point is invalid.
+  /// @param point The point to check.
+  /// @return True if the point is invalid, false otherwise.
+  bool check_invalid_point(const int& channel, const uint16_t& azimuth)
+  {
+    if (!sensor_configuration_->invalid_point_remove) return false;
+    for (size_t i = 0; i < sensor_configuration_->invalid_rings.size(); ++i) {
+      if (
+        sensor_configuration_->invalid_rings.at(i) == channel &&
+        azimuth >= sensor_configuration_->invalid_angles_start.at(i) &&
+        azimuth <= sensor_configuration_->invalid_angles_end.at(i)) {
+        return true;
+      }
+    }
+    return false;
+  }
 };
 
 }  // namespace drivers
--- a/nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp16_decoder.cpp
+++ b/nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp16_decoder.cpp
@@ -273,6 +273,9 @@ void Vlp16Decoder::unpack(const velodyne_msgs::msg::VelodynePacket & velodyne_pa
                 current_point.return_type = return_type;
                 current_point.channel = corrections.laser_ring;
                 current_point.azimuth = rotation_radians_[azimuth_corrected];
+
+                if (check_invalid_point(corrections.laser_ring, azimuth_corrected)) continue;
+
                 current_point.elevation = sin_vert_angle;
                 auto point_ts = block_timestamp - scan_timestamp_ + point_time_offset;
                 if (point_ts < 0)
--- a/nebula_common/include/nebula_common/velodyne/velodyne_common.hpp
+++ b/nebula_common/include/nebula_common/velodyne/velodyne_common.hpp
@@ -19,6 +19,10 @@ struct VelodyneSensorConfiguration : SensorConfigurationBase
   uint16_t rotation_speed;
   uint16_t cloud_min_angle;
   uint16_t cloud_max_angle;
+  bool invalid_point_remove;
+  std::vector<int64_t> invalid_rings;
+  std::vector<int64_t> invalid_angles_start;
+  std::vector<int64_t> invalid_angles_end;
 };
 /// @brief Convert VelodyneSensorConfiguration to string (Overloading the << operator)
 /// @param os
--- a/nebula_common/include/nebula_common/velodyne/velodyne_common.hpp
+++ b/nebula_common/include/nebula_common/velodyne/velodyne_common.hpp
@@ -11,6 +11,14 @@ namespace nebula
 {
 namespace drivers
 {
+
+/// @brief Invalid region on the cloud to be removed
+struct InvalidRegion {
+  int ring;
+  uint16_t start;
+  uint16_t end;
+};
+
 /// @brief struct for Velodyne sensor configuration
 struct VelodyneSensorConfiguration : SensorConfigurationBase
 {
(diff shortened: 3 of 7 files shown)


amc-nu:
This looks like an interesting feature.
However, I think this should be implemented as a composable node for any current and future lidar model.

mebasoglu:
Hello @drwnz , this PR is ready for a review.

README.md
```diff
@@ -214,14 +214,34 @@ Parameters shared by all supported models:
...
-| cloud_max_angle  | uint16 | 359      | degrees [0, 360] | FoV end angle                 |
+| Parameter              | Type   | Default  | Accepted values  | Description                      |
+|------------------------|--------|----------|------------------|----------------------------------|
+| frame_id               | string | velodyne |                  | ROS frame ID                     |
+| calibration_file       | string |          |                  | LiDAR calibration file           |
+| min_range              | double | 0.3      | meters, >= 0.3   | Minimum point range published    |
+| max_range              | double | 300.0    | meters, <= 300.0 | Maximum point range published    |
+| cloud_min_angle        | uint16 | 0        | degrees [0, 360] | FoV start angle                  |
+| cloud_max_angle        | uint16 | 359      | degrees [0, 360] | FoV end angle                    |
+| invalid_point_remove   | bool   | false    | true, false      | Enable ring based filter*        |
```
xmfcx, line 225:
```suggestion
| enable_ring_section_filter   | bool   | false    | true, false      | Enable ring section filter*        |
```

We could name this feature Ring Section Filter, what do you think?

README.md
```diff
@@ -214,14 +214,34 @@ Parameters shared by all supported models:
...
+| invalid_regions        | string |          |                  | Invalid point regions to remove* |
```
xmfcx, line 226:
```suggestion
| excluded_ring_sectors        | string |          |                  | 	Identifies and prevents specific ring sectors from being included in the point cloud based on ring ID, start angle, and end angle. |
```

README.md
```diff
@@ -214,14 +214,34 @@ Parameters shared by all supported models:
...
+
+*`invalid_point_remove` activates the ring based filter which removes points if they are within specified angle ranges. 
+
+*The format for an invalid region is [ring_id, start_angle, end_angle]
+
+*Angles are given in degrees and multiplied by 100. For instance, 34.44 degrees is represented as 3444. 
+
+*Invalid regions are specified as a string containing a list of invalid regions. Ensure that you have quotation marks to make it string. For example:
```
xmfcx, line 234:
```suggestion
- `enable_ring_section_filter` toggles a ring-based filter to remove points located within predefined angle ranges.
- Specify an excluded section using the format `[ring_id, start_angle, end_angle]`.
- Angles must be in degrees, scaled by a factor of 100. For example, represent `34.44` degrees as `3444`.
- It's possible to define multiple excluded regions for the same ring, allowing for versatile filtering configurations.
- Define excluded regions as a string containing a list of such regions, enclosed in quotation marks. For instance:
```

README.md
```diff
@@ -214,14 +214,34 @@ Parameters shared by all supported models:
...
+```xml
+<node pkg="nebula_ros" exec="velodyne_driver_ros_wrapper_node"
+          name="velodyne_cloud" output="screen">
+        ...
+        <param name="invalid_point_remove" value="true"/>
+        <param name="invalid_regions" value="'[[0, 3500, 6900], [1, 3400, 6500], [2, 3200, 4600], [3, 3200, 4600]]'"/>
+    </node>
+```
+
+*Multiple invalid regions are possible for the same ring.
```
xmfcx, line 244:
```suggestion
```

README.md
```diff
@@ -214,14 +214,34 @@ Parameters shared by all supported models:
...
+
+*Angles are given in degrees and multiplied by 100. For instance, 34.44 degrees is represented as 3444. 
+
+*Invalid regions are specified as a string containing a list of invalid regions. Ensure that you have quotation marks to make it string. For example:
+```xml
+<node pkg="nebula_ros" exec="velodyne_driver_ros_wrapper_node"
+          name="velodyne_cloud" output="screen">
+        ...
+        <param name="invalid_point_remove" value="true"/>
+        <param name="invalid_regions" value="'[[0, 3500, 6900], [1, 3400, 6500], [2, 3200, 4600], [3, 3200, 4600]]'"/>
```
xmfcx, line 240:
```suggestion
        <param name="invalid_regions" value="'[[0, 3500, 6900], [1, 3400, 6500], [2, 3200, 4600], [3, 3200, 4600]]'"/>
```

Why do we need the additional single quotes?

mebasoglu: Hello, since ROS 2 doesn't support setting parameters as a list of lists, Boost Property Tree library has been used.
`invalid_regions` is assumed as a string and parsed with the library. To make it string, I have added another set of quotes.

nebula_ros/include/nebula_ros/velodyne/velodyne_decoder_ros_wrapper.hpp
```diff
@@ -15,6 +15,9 @@
 #include <velodyne_msgs/msg/velodyne_packet.hpp>
 #include <velodyne_msgs/msg/velodyne_scan.hpp>
 
+#include "boost/property_tree/json_parser.hpp"
+#include "boost/property_tree/ptree.hpp"
```
xmfcx, line 19:
```suggestion
#include <nlohmann/json.hpp>
```

Let's use this instead, boost is too heavy for this task.

`<depend>nlohmann-json-dev</depend>` can be added to the `package.xml` file and all should work fine.

And the rest of the code will be much cleaner and also will be error checked:

```cpp
// Assuming `nlohmann::json` namespace is available as `json`
void parseInvalidRegions(const std::string& regions_json) {
    // Parse the JSON string
    auto invalid_regions_json = json::parse(regions_json);

    // Clear existing regions
    sensor_configuration_.invalid_regions.clear();

    // Iterate over the parsed JSON array
    for (const auto& region : invalid_regions_json) {
        // Extract the ring number, start angle, and end angle from each sub-array
        int ring_number = region[0];
        uint16_t start = region[1];
        uint16_t end = region[2];

        // Add the InvalidRegion to the map under the corresponding ring number
        sensor_configuration_.invalid_regions[ring_number].emplace_back(InvalidRegion{start, end});
    }
}
```

mebasoglu:
Hello @drwnz @mojomex , could you please review this PR? Maybe we can merge it before https://github.com/tier4/nebula/pull/138

nebula_decoders/include/nebula_decoders/nebula_decoders_velodyne/decoders/velodyne_scan_decoder.hpp
```diff
@@ -182,6 +182,19 @@ class VelodyneScanDecoder
   virtual void reset_pointcloud(size_t n_pts, double time_stamp) = 0;
   /// @brief Resetting overflowed point cloud buffer
   virtual void reset_overflow(double time_stamp) = 0;
+
```
drwnz, line 185: This check has to be calculated with a search in the excluded regions parameters for every point, which is causing the performance overhead. We suggest pre-computing the binary value for each azimuth and channel value to make a map, allowing a single conditional to check if the point is valid or not.

nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp16_decoder.cpp
```diff
@@ -237,13 +237,20 @@ void Vlp16Decoder::unpack(const velodyne_msgs::msg::VelodynePacket & velodyne_pa
...
                 (azimuth_corrected >= sensor_configuration_->cloud_min_angle * 100 &&
                  azimuth_corrected <= sensor_configuration_->cloud_max_angle * 100 &&
-                 sensor_configuration_->cloud_min_angle < sensor_configuration_->cloud_max_angle) ||
+                 sensor_configuration_->cloud_min_angle < sensor_configuration_->cloud_max_angle);
+
+              const bool is_outside_max_min_angle =
                 (sensor_configuration_->cloud_min_angle > sensor_configuration_->cloud_max_angle &&
                  (azimuth_corrected <= sensor_configuration_->cloud_max_angle * 100 ||
-                  azimuth_corrected >= sensor_configuration_->cloud_min_angle * 100))) {
+                  azimuth_corrected >= sensor_configuration_->cloud_min_angle * 100));
```
drwnz, line 248: Suggestion: as mentioned in the WG meeting, it would be nice to encapsulate this code into a filter class to make it clearer what function is being performed in the decoder, and also reduce repeated code in each decoder.

drwnz (changes requested):
Thank you for the PR, and apologies for the delay in review!
We would really like to apply this methodology to Hesai decoders as well, so if you can look at the suggested changes, we will proceed quickly to review. Mostly, if we are to adopt throughout multiple sensor manufacturers and models, we need to make sure to minimize the performance cost.

mojomex (changes requested):
@mebasoglu Thank you for your contribution!
I'm sorry for responding _that_ late :bow: but here is my review.

The feature itself looks great and is a nice addition to Nebula but I would like you to refactor the filter code into its own isolated class as much as possible to properly separate the sensor features from Nebula features. Please respond to the comments I left above.

Thank you!

nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp16_decoder.cpp
drwnz: Sorry this comment is on the wrong part of the code, a mistake on my part! Will delete

mebasoglu:
Hello @drwnz @mojomex, unfortunately I couldn't take a look at the PR since your review. It seems we aren't able to work on this right now. I wonder if it's all right for you to continue with this issue?

mojomex:
@mebasoglu As we want to implement something similar for Hesai sensors too, I think we can pick up where you left off. Not sure if it will happen this week, but definitely soon!

mojomex:
Closing in favor of 
* #250

PR #72, opened 2023-09-11
fix(hesai_hw_interface): hesai interface

amc-nu:
## PR Type

- Improvement

## Related Links

https://tier4.atlassian.net/browse/RT1-3200
https://tier4.atlassian.net/browse/RT1-3111

## Description

Improves reliability of Hesai HW Interface.:
- Allows `sensor_setup:=False`
- Removes crashing when sensor not available or timeout.

Includes the following features:

* Gets LidarCalibration from sensor
* Skips GetInventory if unable to connect.
* Fixes Hesai monitor.
* Check calibration file format AT128.
* Saves calibration data from sensor.
* Fallback to offline mode AT128
* Checks Return Mode AT128

## CI Checks


drwnz (commented):
When restarting the driver multiple times with the device connected, after the 4th attempt I get this warning:
```[component_container-1] [WARN] [1697789855.726097568] [hesai_hw_monitor_ros_wrapper_node]: diagnostic_updater: No HW_ID was set. This is probably a bug. Please report it. For devices that do not have a HW_ID, set this value to 'none'. This warning only occurs once all diagnostics are OK. It is okay to wait until the device is open before calling setHardwareID.```

This is coming from diagnostic updater - I guess this is related to the TCP connection limitation we discussed? FYI this is on the AT128.
The device still works, and the configuration also seems to stick so not a major issue.

amc-nu:
Yes, exactly. The sensor stops receiving commands.

drwnz:
> Yes, exactly. The sensor stops receiving commands.

It does seem to still correctly configure the sensor though?

amc-nu:
It configures the sensor only during the first attempts. Afterwards the sensor rejects all the connections.

drwnz:
I will merge this PR and we can work on the TCP rejection issue in a different PR.
Thank you for the great work, DAT file download works well.

drwnz (approved):

PR #84, opened 2023-10-30
fix: make AT128 scans align with scan_phase

mojomex:
## PR Type

- Bug Fix

## Related Links

[TIER IV INTERNAL LINK](https://tier4.atlassian.net/l/cp/T1YrUA0Z) -- AT128 `scan_angle` troubleshooting log

## Description

Before this PR, AT128's pointcloud timestamps were not aligned to top of second at the set `scan_angle`.
The refactoring in #42 introduced a wrong calculation for scan cutting in AT128 (subtracted `scan_angle` in output coordinates from the raw azimuth value).
Further, AT128's 3.50.6 software version set `scan_angle` as-is, without converting internally to encoder angles first.
This has been fixed in 3.50.8.

This PR fixes this issue via the following code changes:
* handle `scan_angle` in the respective `angle_corrector_...` implementations, not in `hesai_decoder`
* set the `scan_phase` parameter in the AT128 configuration file to `30.0` (the lowest legal value)
* add AT128 software version requirements to the readme file

## Review Procedure

Run Nebula with AT128, set `scan_angle` to different values `30 <= scan_angle <= 150` and observe the timestamps of the output pointclouds:
```bash
ros2 topic echo --field header.stamp.nanosec /pandar_points 
```
If the second and third digits are always close to `0`, the fix is working.

## Remarks

Because channel corrections can range from around -3 to +3 deg for the same encoder angle, there is a range of output angles associated with the encoder angle. Currently, the scan cutting/angle conversion code ignores this. It is a design decision whether to align to the earliest channel to hit `scan_angle` or to align channel-agnostically (without factoring in correction). The latter ends up roughly in the mean of all channels.

Here is an excerpt from [the extensive analysis](https://tier4.atlassian.net/l/cp/T1YrUA0Z), showing `scan_angle = 90 deg`. Black is aligned to `0.X000... s`, yellow is maximally unaligned (`0.Y999... s`). Channel 0 is the innermost ring, channel 127 the outermost. One can see that the points at the same output angle are not all aligned to ToS.
image

- [x] Assign PR to reviewer

## CI Checks


nebula_tests/hesai/hesai_ros_decoder_test_main.cpp
```diff
@@ -31,7 +31,8 @@ const nebula::ros::HesaiRosDecoderTestParams TEST_CONFIGS[6] = {
   },
   {
     .sensor_model = "PandarAT128",
-    .return_mode = "LastStrongest",
+    .return_mode = "Dual",
+    .scan_phase = 152,
```
mojomex, line 35: We would have to re-record new unit test reference data for AT128 with known `scan_phase` setting to make this a nice number. Because it is unknown to me at the moment, I adjusted it to make the test pass.
Because there is a certain leeway anyway with how to sync to `scan_phase` (see the remarks section of this PR), I don't see a big issue with leaving in this somewhat specific number.

knzo25:
@mojomex 
FYI, internal links in public repositories are usually presented as TIER IV INTERNAL LINK
:+1:

drwnz (changes requested):
Great stuff, and ready to merge - but do you mind removing 14b43c0c0e9b69fc941e422a246e1169017fff73 to make merging as simple as possible?

mojomex:
Done :+1:

drwnz (approved):
:partying_face:

PR #102, opened 2023-12-04
perf(hw_interface): write packet data directly into the message

xmfcx:
## PR Type

- Improvement

## Related Links

None.

## Description

Before this change, in the driver, every time a UDP packet is received, it is first written to a `std::array` object, then copied into the message.
With this change, the UDP packets are directly read into the message, avoiding the overhead.

Their types are the same so it won't cause an issue.

Small note: also made the buffer size variable `const` since it doesn't change over time.

## Review Procedure

You could:
- double check the types.
- test on pcap data.

- [x] Assign PR to reviewer

## CI Checks


mojomex:
Confirmed working on VLP32

drwnz (approved):
Thanks @xmfcx, looks good!

PR #111, opened 2023-12-28
fix(velodyne_decoder): overflow handling in vls128

knzo25:
The timestamps in the vls128 were still relative to the last pointcloud, which was producing some errors Additionally, when entire scans were dropped, the overflow was not being handled correctly (this bug was also present in the awf velodyne driver)

## PR Type

- Bug Fix

## Related Links
[TIER IV INTERNAL LINK (discussion)](https://star4.slack.com/archives/C057JDFS5M2/p1703655034662449)
[TIER IV INTERNAL LINK (footage)](https://drive.google.com/drive/folders/1sB3iTCVo9Z4ysu-HwtBLWnPHRN3Vmk6_?usp=drive_link)

## Description
2 bugs were discovered and addressed related to the handling of time stamps during pointcloud overflow between scans.
Context: 
 - The lidar scan has an azimuth value (the origin is defined by the sensor) and usually a pointcloud would start at 0 degrees and end in 360 degrees. 
 - However, to synchronize multiple lidars and cameras we can actually make pointclouds  starting from a different angle.
 - Since the communication with the sensor is via "small" packets instead of a single packet per physical scan (360 degrees) we can decide how to assemble the packets into a pointcloud (e.g, start making a pointcloud starting from sensor angle = 30 degrees).
 - However, since each packet contains several points with different azimuth values, we need to handle one packet in two poinclouds (we call this concept overflow). Example: a certain packet contains points both lower and higher than a desired azimuth threshold. Lower values are added to the current pointcloud, whereas the higher ones need to be handled by the next one.

Nebula-only problem:
 - When a scan message arrives at the decoder, the absolute time stamp gets taken from the first block of the first package in the scan message. In nebula (velodyne_points), each individual point has a relative time stamp, and they are relative to the previously stated absolute time stamp.
 - However, the points that were determined to be part of the next pointcloud (overflow) currently retain their relative timestamp. Since they are the last points of the scan, they actually have relative times close to 100ms (the highest value, since physical scans run at 10hz).
 - When the next scan message arrives, the process is repeated once again, but this time, the first points of the pointcloud will have a very high relative time, followed by points with relative time close to 0 and increasing to almost 100ms (there is a discontinuity).

Although only the non-overflow points are OK, if we use algorithms that integrate timestamps from all the points starting from the beginning (overflow points), it will cause big errors. This was particularly observed in the distortion correction algorithm (see the provided videos)

The solution simply consists of using as global timestamp the stamp of the overflow points (some conversions were required).

Nebula & old vls driver problem:
 - In the vehicle, due to how pointcloud containers work, "there are no dropped packages". All packets from the lidar end up being converted into pointclouds.
 - However, when recording and replaying rosbags, some entire scans (aggregation of sensor packets corresponding to one physical 360 degree scan) can be missing.
 - In this case, even using the previews fix, errors still occur, since the overflow packets are used to determine the global time stamp, but since the points following the overflow are missing, the points after the overflow corresponds to the ones of the next physical scan, which makes a total time difference between the first and last point of the pointcloud of about 200ms (two scans).

For this problem, the proposed solution is that we first detect dropped packages using a rough stamp criteria, and then just ignore the last pointcloud's overflow when necessary (since there are abut 600 packages per scan, we are only loosing 360 * 1 / 600 degrees of information)

## Review Procedure
A good way to check that the timestamps are being computed correctly, is comparing the raw_pointcloud with the output of the distortion corrector of autoware.

For example, it would be convenient to compare the behavior of the output of the distortion correction before/after this PR
(additionally, the fix for the original vls implementation also serves as a good baseline https://github.com/knzo25/awf_velodyne/tree/fix/drop_packages_fix )

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
+++ b/nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
@@ -116,16 +116,17 @@ void Vls128Decoder::reset_overflow(double time_stamp)
   }
 
   // Add the overflow buffer points
-  for (size_t i = 0; i < overflow_pc_->points.size(); i++) {
-    auto overflow_point = overflow_pc_->points[i];
+  while (overflow_pc_->points.size() > 0) {
+    auto overflow_point = overflow_pc_->points.back();
 
     // The overflow points had the stamps from the previous pointcloud. These need to be changed to
     // be relative to the overflow's packet timestamp
     double new_timestamp_seconds =
       scan_timestamp_ + 1e-9 * overflow_point.time_stamp - last_block_timestamp_;
-    overflow_point.time_stamp = 1e9 * new_timestamp_seconds;
+    overflow_point.time_stamp = new_timestamp_seconds < 0.0 ? 0.0 : 1e9 * new_timestamp_seconds;
 
     scan_pc_->points.emplace_back(overflow_point);
+    overflow_pc_->points.pop_back();
   }
 
   // When there is overflow, the timestamp becomes the overflow packets' one
--- a/nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
+++ b/nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
@@ -36,37 +36,42 @@ Vls128Decoder::Vls128Decoder(
   }
   // timing table calculation, from velodyne user manual p.64
   timing_offsets_.resize(3);
-  for(size_t i=0; i < timing_offsets_.size(); ++i)
-  {
-    timing_offsets_[i].resize(17); // 17 (+1 for the maintenance time after firing group 8)
+  for (size_t i = 0; i < timing_offsets_.size(); ++i) {
+    timing_offsets_[i].resize(17);  // 17 (+1 for the maintenance time after firing group 8)
   }
   double full_firing_cycle_s = 53.3 * 1e-6;
   double single_firing_s = 2.665 * 1e-6;
   double offset_packet_time = 8.7 * 1e-6;
   // compute timing offsets
-  for (size_t x = 0; x < timing_offsets_.size(); ++x){
-    for (size_t y = 0; y < timing_offsets_[x].size(); ++y){
+  for (size_t x = 0; x < timing_offsets_.size(); ++x) {
+    for (size_t y = 0; y < timing_offsets_[x].size(); ++y) {
       double sequence_index, firing_group_index;
       sequence_index = x;
       firing_group_index = y;
-      timing_offsets_[x][y] = (full_firing_cycle_s * sequence_index) + (single_firing_s * firing_group_index) - offset_packet_time;
+      timing_offsets_[x][y] = (full_firing_cycle_s * sequence_index) +
+                              (single_firing_s * firing_group_index) - offset_packet_time;
     }
   }
 }
 
-bool Vls128Decoder::hasScanned() { return has_scanned_; }
+bool Vls128Decoder::hasScanned()
+{
+  return has_scanned_;
+}
 
 std::tuple<drivers::NebulaPointCloudPtr, double> Vls128Decoder::get_pointcloud()
 {
   double phase = angles::from_degrees(sensor_configuration_->scan_phase);
   if (!scan_pc_->points.empty()) {
     auto current_azimuth = scan_pc_->points.back().azimuth;
-    auto phase_diff = static_cast<size_t>(angles::to_degrees(2*M_PI + current_azimuth - phase)) % 360;
+    auto phase_diff =
+      static_cast<size_t>(angles::to_degrees(2 * M_PI + current_azimuth - phase)) % 360;
     while (phase_diff < M_PI_2 && !scan_pc_->points.empty()) {
       overflow_pc_->points.push_back(scan_pc_->points.back());
       scan_pc_->points.pop_back();
       current_azimuth = scan_pc_->points.back().azimuth;
-      phase_diff = static_cast<size_t>(angles::to_degrees(2*M_PI + current_azimuth - phase)) % 360;
+      phase_diff =
+        static_cast<size_t>(angles::to_degrees(2 * M_PI + current_azimuth - phase)) % 360;
     }
     overflow_pc_->width = overflow_pc_->points.size();
     scan_pc_->width = scan_pc_->points.size();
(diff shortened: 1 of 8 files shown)


nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
```diff
@@ -75,24 +80,56 @@ std::tuple<drivers::NebulaPointCloudPtr, double> Vls128Decoder::get_pointcloud()
...
 {
+  if (overflow_pc_->points.size() == 0) {
+    scan_timestamp_ = -1;
+    overflow_pc_->points.reserve(max_pts_);
+    return;
+  }
+
+  // Compute the absolute time stamp of the last point of the overflow pointcloud
+  const double last_overflow_time_stamp =
+    scan_timestamp_ + 1e-9 * overflow_pc_->points.back().time_stamp;
```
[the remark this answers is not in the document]
bgilby59: Or just add a comment that there is a ns to s conversion?

nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
```diff
@@ -75,24 +80,56 @@ std::tuple<drivers::NebulaPointCloudPtr, double> Vls128Decoder::get_pointcloud()
...
   // Add the overflow buffer points
   for (size_t i = 0; i < overflow_pc_->points.size(); i++) {
-    scan_pc_->points.emplace_back(overflow_pc_->points[i]);
+    auto overflow_point = overflow_pc_->points[i];
+
+    // The overflow points had the stamps from the previous pointcloud. These need to be changed to
+    // be relative to the overflow's packet timestamp
+    double new_timestamp =
+      scan_timestamp_ + 1e-9 * overflow_point.time_stamp - last_block_timestamp_;
+    overflow_point.time_stamp = 1e-9 * new_timestamp;
```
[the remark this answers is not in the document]
knzo25: Thanks for the catch !
Fixed in [cf0c795](https://github.com/tier4/nebula/pull/111/commits/cf0c79527f7c98e947b81c5100ceec9ed0ec7eb0)

nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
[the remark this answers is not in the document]
knzo25: While I agree with improving readability, this conversion is all over the place in nebula, so for now I would prefer to leave it as is.

knzo25:
@bgilby59 
Regarding the additional problems we experienced the other day, [9091b2f](https://github.com/tier4/nebula/pull/111/commits/9091b2fe40afbbb4fe8c812490eef4ab1faa102d) should fix those.

nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
```diff
@@ -116,16 +116,17 @@ void Vls128Decoder::reset_overflow(double time_stamp)
   }
 
   // Add the overflow buffer points
-  for (size_t i = 0; i < overflow_pc_->points.size(); i++) {
-    auto overflow_point = overflow_pc_->points[i];
+  while (overflow_pc_->points.size() > 0) {
+    auto overflow_point = overflow_pc_->points.back();
```
bgilby59, line 120: Why did you change the for loop to a while loop? Looks fine, but I'm just curious about the motivation to make the change.

knzo25: Sorry, forgot to mention it before.
If you look at the way the overflow_pc if constructed, the order of the points is reversed.
In the implementation prior to [9091b2f](https://github.com/tier4/nebula/pull/111/commits/9091b2fe40afbbb4fe8c812490eef4ab1faa102d) the new order was kept and copied into the new pc. That is, the output pc was reversed.

While checking the timestamps of the pc I realized that the timestamps were not in ascending order, which led me to find the error. That is way, by reversing the order again while moving the overflow into the new pc, the "correct" order is preserved

bgilby59: Ah, I see now. The previous implementation was using `emplace_back()`. Thanks!

nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
```diff
@@ -75,24 +80,57 @@ std::tuple<drivers::NebulaPointCloudPtr, double> Vls128Decoder::get_pointcloud()
...
-  for (size_t i = 0; i < overflow_pc_->points.size(); i++) {
-    scan_pc_->points.emplace_back(overflow_pc_->points[i]);
+  while (overflow_pc_->points.size() > 0) {
+    auto overflow_point = overflow_pc_->points.back();
+
+    // The overflow points had the stamps from the previous pointcloud. These need to be changed to
+    // be relative to the overflow's packet timestamp
+    double new_timestamp_seconds =
+      scan_timestamp_ + 1e-9 * overflow_point.time_stamp - last_block_timestamp_;
+    overflow_point.time_stamp = new_timestamp_seconds < 0.0 ? 0.0 : 1e9 * new_timestamp_seconds;
```
bgilby59, line 126:
```suggestion
    overflow_point.time_stamp = static_cast<uint32_t>(new_timestamp_seconds < 0.0 ? 0.0 : 1e9 * new_timestamp_seconds);
```

nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
```diff
@@ -75,24 +80,57 @@ std::tuple<drivers::NebulaPointCloudPtr, double> Vls128Decoder::get_pointcloud()
...
+    scan_timestamp_ = -1;
+    overflow_pc_->points.reserve(max_pts_);
+    return;
+  }
+
+  // Compute the absolute time stamp of the last point of the overflow pointcloud
+  const double last_overflow_time_stamp =
+    scan_timestamp_ + 1e-9 * overflow_pc_->points.back().time_stamp;
+
+  // Detect cases where there in an unacceptable time difference between the last overflow point and
```
bgilby59, line 109: Typo fix:
```suggestion
  // Detect cases where there is an unacceptable time difference between the last overflow point and
```

bgilby59 (approved):
Looks good. Tested on my end with 3 rosbags with XX1 data. The rosbags used and videos of results can be found at the below link:

[TIER IV INTERNAL LINK](https://drive.google.com/drive/folders/1DtnhiGz5xRRozB9l0LVhHEL9ceG5xh_4)

PR #117, opened 2024-02-04
build: make #include <angles/angles.h> internal to cpp file

felixf4xu:
## PR Type

- Improvement

## Description

before this PR, `angles/angles.h` is included in a public header file of `velodyne_scan_decoder.hpp`, so that any project that uses this repo needs to include `angles` as well.

After this PR, `angles/angles.h` is moved to .cpp files, so it's internal to this repo. Other project that uses this repo does not need to find and include `angles`.

This can reduce compile time for both colcon/cmake and gcc.

Another minor change is the 
```
#if defined(ROS_DISTRO_FOXY) || defined(ROS_DISTRO_GALACTIC)
#include <angles/angles.h>  //Galactic
#else
#include <angles/angles/angles.h>  //Humble
#endif
```
which is not necessary if `angles` is installed according to https://github.com/ros/rosdistro/blob/master/humble/distribution.yaml#L403

Many other ros2 projects are not using `#if/#endif` for humble like https://github.com/ros-planning/navigation2/blob/humble/nav2_controller/plugins/pose_progress_checker.cpp#L20
## Review Procedure

build this project

- [ ] Assign PR to reviewer

- [x] All open points are addressed and tracked via issues or tickets

## CI Checks

### DIFF
--- a/nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp16_decoder.cpp
+++ b/nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp16_decoder.cpp
@@ -3,6 +3,8 @@
 #include <cmath>
 #include <utility>
 
+#include <angles/angles.h> 
+
 namespace nebula
 {
 namespace drivers
--- a/nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp32_decoder.cpp
+++ b/nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp32_decoder.cpp
@@ -3,6 +3,8 @@
 #include <cmath>
 #include <utility>
 
+#include <angles/angles.h> 
+
 namespace nebula
 {
 namespace drivers
--- a/nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
+++ b/nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
@@ -3,6 +3,8 @@
 #include <cmath>
 #include <utility>
 
+#include <angles/angles.h> 
+
 namespace nebula
 {
 namespace drivers
--- a/nebula_decoders/CMakeLists.txt
+++ b/nebula_decoders/CMakeLists.txt
@@ -14,6 +14,7 @@ if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     add_compile_options(-Wall -Wextra -Wpedantic -Wunused-function)
 endif ()
 
+find_package(angles REQUIRED)
 find_package(PCL REQUIRED)
 find_package(pcl_conversions REQUIRED)
 find_package(yaml-cpp REQUIRED)
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_velodyne/decoders/velodyne_scan_decoder.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_velodyne/decoders/velodyne_scan_decoder.hpp
@@ -12,12 +12,6 @@
 #include <string>
 #include <vector>
 
-#if defined(ROS_DISTRO_FOXY) || defined(ROS_DISTRO_GALACTIC)
-#include <angles/angles.h>  //Galactic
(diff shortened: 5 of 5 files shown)


nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp16_decoder.cpp
```diff
@@ -3,6 +3,8 @@
 #include <cmath>
 #include <utility>
 
+#include <angles/angles.h>
```
knzo25, line 6:
```suggestion
#if defined(ROS_DISTRO_FOXY) || defined(ROS_DISTRO_GALACTIC)
#include <angles/angles.h>  //Galactic
#else
#include <angles/angles/angles.h>  //Humble
#endif
```

nebula_decoders/src/nebula_decoders_velodyne/decoders/vlp32_decoder.cpp
```diff
@@ -3,6 +3,8 @@
 #include <cmath>
 #include <utility>
 
+#include <angles/angles.h>
```
knzo25, line 6:
```suggestion
#if defined(ROS_DISTRO_FOXY) || defined(ROS_DISTRO_GALACTIC)
#include <angles/angles.h>  //Galactic
#else
#include <angles/angles/angles.h>  //Humble
#endif
```

nebula_decoders/src/nebula_decoders_velodyne/decoders/vls128_decoder.cpp
```diff
@@ -3,6 +3,8 @@
 #include <cmath>
 #include <utility>
 
+#include <angles/angles.h>
```
knzo25, line 6:
```suggestion
#if defined(ROS_DISTRO_FOXY) || defined(ROS_DISTRO_GALACTIC)
#include <angles/angles.h>  //Galactic
#else
#include <angles/angles/angles.h>  //Humble
#endif
```

knzo25:
@felixf4xu 
Thanks for the contribution

felixf4xu:
@knzo25 Thanks for the comment.

As to the code, could you verify again that in `humble`, the header file path is `<angles/angles/angles.h>`? 

Actually I was building autoware on ros2 humble and I can't find the header file at the path of `<angles/angles/angles.h>`, and I had to change it to `<angles/angles.h>`. In that case, there will no difference for `humble` or `foxy`.

As an example, the `rclcpp.hpp` full path is  `/opt/ros/humble/include/rclcpp/rclcpp/rclcpp.hpp` ( 2 `rclcpp` in the path), but when we include it, we only include `<rclcpp/rclcpp.hpp>` (1 `rclcpp` in the path). 

In my ubuntu 22, `angles.h` is at `/opt/ros/humble/include/angles/angles/angles.h` (2 `angles` in the path), but in my case, I have to use `<angles/angles.h>` (1 `angles` in the name) to compile.

knzo25:
Hi, sorry for the earlier response. 
In the past we had issues with some changes in naming, but in this ocassion it seems it is as you mention.

Summary:
Before your PR, we were doing
`#include <angles/angles/angles.h>  //Humble`
In our code, which is is not standard, as the real path of the headers is

```
kenzolobos@pcname:/opt/ros/humble/include/angles/angles$ find /opt/ros/humble/include/angles
/opt/ros/humble/include/angles
/opt/ros/humble/include/angles/angles
/opt/ros/humble/include/angles/angles/angles.h

```

hence, the includes should be:

`#include <angles/angles.h>`

The reason using our old method still works is because the actual g++ command executed includes the following line 
`-isystem /opt/ros/humble/include`

Which allowed to include headers starting at the general ROS level

In addition to the previous g++ flag, currently the compiler is adding the following:
`-isystem /opt/ros/humble/include/angles` which enables standard ros includes.

As you can see from the `find` command above the package angles itself does not contain anything other than the headers, so it may be that in the past the correct compiler flag was not being properly added so the developers at the time fell back to using paths starting at the ros level.

I will be deleting my comments and approving the PR. Again, thanks for the contribution !

knzo25 (approved):
LGTM !

PR #125, opened 2024-03-11
feat(hesai): add TSN settings for Hesai sensors (OT128) 

drwnz:
## PR Type

- New Feature

## Description

Hesai OT128 has different handling of PTP settings, as it only allows 802.1AS Automotive and allows setting of switch type (TSN or not). This in turn changes PTP from the default P2P to use E2E when the switch type is changed to non-TSN.

## Review Procedure

- Connect OT128 sensor
- Configure sensor with 802.1AS Automotive via launchfile, and set `ptp_switch_type` to `TSN` or `NON_TSN`
- Confirm PTP operation, and also check that the delay measurement mechanism is correctly configured

## Remarks

ToDo:
- [ ] Tests to clear build and test if required

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
+++ b/nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
@@ -2336,6 +2336,12 @@ Status HesaiHwInterface::SetPtpConfig(
   buf_vec.emplace_back((len >> 8) & 0xff);
   buf_vec.emplace_back((len >> 0) & 0xff);
 
+  if (sensor_configuration_->sensor_model == SensorModel::HESAI_PANDAR128_E4X) {
+    if (profile != static_cast<int>(PtpProfile::IEEE_802_1AS_AUTO)) {
+      return Status::SENSOR_CONFIG_ERROR;
+    }
+    profile = 3; // OT128 has a different definition of PTP profile
+  }
   buf_vec.emplace_back((profile >> 0) & 0xff);
   buf_vec.emplace_back((domain >> 0) & 0xff);
   buf_vec.emplace_back((network >> 0) & 0xff);
@@ -2344,7 +2350,7 @@ Status HesaiHwInterface::SetPtpConfig(
     buf_vec.emplace_back((logSyncInterval >> 0) & 0xff);
     buf_vec.emplace_back((logMinDelayReqInterval >> 0) & 0xff);
   }
-  if (profile == 3) {
+  else if (profile == 3) {
     buf_vec.emplace_back((switch_type >> 0) & 0xff);
   }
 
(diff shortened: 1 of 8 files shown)


nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
```diff
@@ -2367,29 +2376,29 @@ Status HesaiHwInterface::SetPtpConfig(
 }
 Status HesaiHwInterface::SetPtpConfig(
   std::shared_ptr<boost::asio::io_context> ctx, int profile, int domain, int network,
-  int logAnnounceInterval = 1, int logSyncInterval = 1, int logMinDelayReqInterval = 0,
-  bool with_run)
+  int switch_type, int logAnnounceInterval = 1, int logSyncInterval = 1,
```
mojomex, line 2379: The default parameters were already there before, but usually all parameters after ones that have defaults, have to have defaults too. We should probably add a default for `with_run` or get rid of all of them.
Also, I'm curious as to why the compiler doesn't give a warning..

nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
```diff
@@ -2336,6 +2336,12 @@ Status HesaiHwInterface::SetPtpConfig(
   buf_vec.emplace_back((len >> 8) & 0xff);
   buf_vec.emplace_back((len >> 0) & 0xff);
 
+  if (sensor_configuration_->sensor_model == SensorModel::HESAI_PANDAR128_E4X) {
+    if (profile != static_cast<int>(PtpProfile::IEEE_802_1AS_AUTO)) {
+      return Status::SENSOR_CONFIG_ERROR;
+    }
+    profile = 3; // OT128 has a different definition of PTP profile
+  }
```
mojomex, line 2344: I just realized, we also need to update the size in the header to 4; maybe move this above `int len = 6;` above and introduce a case `else if (profile == 3) { len = 4; }`

mojomex (changes requested):
Apart from the comment above, the code looks good! Thank you.
I'll try and test it with a real sensor tomorrow.

nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
drwnz: Oops. The reason is that there is no compiler error, is because the `with_run` default is in the header file. This is not very good, we should move all the defaults to the header file. Nice catch.

drwnz: Resolved in https://github.com/tier4/nebula/pull/125/commits/9a28e3493b452158792ae83bf171c9a47efdefe3

nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
drwnz: Resolved in https://github.com/tier4/nebula/pull/125/commits/37d7b0f4a1a6ec2d5bffc2a375489dfaae253830

mojomex (approved):
LGTM!

PR #156, opened 2024-05-31
fix: non steady clock and watchdog

knzo25:
## PR Type

- Bug Fix

## Description
The callback was being called even though there were frequent updates to the watchdog.
Printing the stamps, the latest stamp was newer than now. Ros time is not steady and it seems ptp affecting the watchdog's time

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/include/nebula_ros/common/watchdog_timer.hpp
+++ b/nebula_ros/include/nebula_ros/common/watchdog_timer.hpp
@@ -36,7 +36,9 @@ private:
   {
     uint64_t now_ns = node_.get_clock()->now().nanoseconds();
 
-    bool is_late = (now_ns - last_update_ns_) > expected_update_interval_ns_;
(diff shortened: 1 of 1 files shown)


nebula_ros/include/nebula_ros/common/watchdog_timer.hpp
```diff
@@ -36,7 +36,9 @@ class WatchdogTimer
   {
     uint64_t now_ns = node_.get_clock()->now().nanoseconds();
 
-    bool is_late = (now_ns - last_update_ns_) > expected_update_interval_ns_;
+    bool is_late = (last_update_ns_ > now_ns)
```
mojomex, line 39:
```suggestion
    // As the clock is not steady, the update timestamp can be newer than clock.now().
    // Define that edge-case as not being late too.
    bool is_late = (last_update_ns_ > now_ns)
```

PR #164, opened 2024-06-13
feat(nebula_ros): force fp parameters for numerics

amadeuszsz:
## PR Type

- Improvement

## Description
JSON schema does not support floating point numbers  directly. Developers can define only integer or numeric type. Therefore, future unit test based on `check-jsonschema` library will pass integer parameter which were defined as numeric (floating point from ROS node perspective). In that case, dynamic parameter declaration for floating point parameters is required.

Idea:
```
if parameter<double> is double -> declare as usual
if parameter<double> is integer -> cast to double
```

- [ ] Assign PR to reviewer

## CI Checks


amadeuszsz:
@mojomex

mojomex:
Closed in favor of #170

PR #173, opened 2024-06-27
feat(hesai): software-based precise FoV limit & early publishing

### LINKED ISSUE #129: AT128 HW interface crashes on GetLidarRange
When launching Nebula when connected to AT128, the HW interface crashes with a segfault in GerLidarRange.
The AT128 TCP protocol does not support request code `0x23` (Get Azimuth FOV) and sends an error response, which Nebula does not check. Instead it assumes the payload is larger than the error response, resulting in a segfault.

mojomex:
## PR Type

- New Feature

## Description

This PR introduces software-based FoV limits and early publishing for Hesai sensors.

### Software-based FoV
Previously, configuring `cloud_min_angle` and `cloud_max_angle` would set the sensor's hardware filter to those values.
However, the hardware filter is using **raw azimuths**.
This lead to the scan having jagged edges (points missing in some channels, too many points in others).
For certain models, this affected areas of up to ~10deg on both edges of the FoV.

This PR improves upon this in the following way:
1. User sets `cloud_min_angle` and `cloud_max_angle` to the exact FoV they want the output pointcloud to have
2. Nebula retrieves sensor calibration file and computes appropriately over-sized* hardware-FoV
3. Nebula sends hardware-FoV to sensor (or validates it if `setup_sensor == false`)
4. The Nebula decoder cuts the scan at the exact edges defined via the parameters above

### Early publishing
Previously, a scan was published once the packet **after** it was received. This was done because this made the scan completion check easier (could compare 2 azimuths according to some scan completion criterion).
For scans of 360deg, this is not a problem, as there is no time gap between the end of one scan and the beginning of the next.
However, for FoVs < 360deg, this approach increases latency unnecessarily: for an FoV of 180deg and 1 full rotation every 100ms, the additional latency of the finished pointcloud just "sitting there" is 50ms.

This PR improves upon this in the following way:
1. Swap the order of decoding and scan completion check: decode first, then check for completion
2. Check if the current packet completed the current scan, as opposed to whether the current packet is the first belonging to the next scan

### Other

For testing this PR with AT128, I had to fix the LidarRange issue:
* fixes #129 

This was done by checking the sensor model before sending the LidarRange command.

## Review Procedure

Check for the following FoVs
* 0 - 360 deg
* 90 - 270 deg
* 270 - 90 deg
if the pointcloud in Rviz has clean edges at those angles for a 360 deg sensor such as OT128.

Check AT128 in the same way but with FoVs inside [30; 150].

## Remarks

These two improvements have been combined into one PR as they are closely related and would require extra work to function independently.

- [x] Assign PR to reviewer

- [x] (Optional) Unit tests have been written for new behavior

## CI Checks


mojomex:
# :construction: Evaluation

## :green_circle: 360deg rotational sensors

| FoV Setting (min, max) | Resulting pointcloud |
|--------|--------|
| 0, 360 |  |
| 90, 270 |  |
| 270, 90 |  |
| 29, 30 |  |
| 360, 0 |  | 

## :green_circle: AT128 (120deg rotational sensor)

| FoV Setting (min, max) | Resulting pointcloud |
|--------|--------|
| 0, 360 |  |
| 30, 150 |  |
| 60, 120 |  |
| 89, 90 |  |

drwnz (changes requested):
This was tested with 40P with the following issue:
- Setting cut angle out side of the range of FoV start and end, resulted in no error and also pointcloud output.
- Even if that was the case and the "cut" happens in the area outside of the FoV (which is kind of uneccesary but logically possible) and the actual observed point cloud is correct in terms of min and max angles.
- Example: cut 10, min 180, max 360: no warnings, output pointcloud is full 180 degree segment.

nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/angle_corrector_correction_based.hpp
```diff
@@ -79,31 +179,62 @@ class AngleCorrectorCorrectionBased : public AngleCorrector<HesaiCorrection>
     auto elevation =
```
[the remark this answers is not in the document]
mojomex: Changed to `int32_t`.

nebula_common/include/nebula_common/hesai/hesai_common.hpp
```diff
@@ -357,7 +377,7 @@ struct HesaiCorrection : public HesaiCalibrationConfigurationBase
   /// @param ch The channel id
   /// @param azi The precision azimuth in (0.01 / 256) degree unit
   /// @return The azimuth adjustment in 0.01 degree unit
-  int8_t getAzimuthAdjustV3(uint8_t ch, uint32_t azi) const
+  [[nodiscard]] int8_t getAzimuthAdjustV3(uint8_t ch, uint32_t azi) const
```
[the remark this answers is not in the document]
mojomex: Agreed :+1:

nebula_common/include/nebula_common/hesai/hesai_common.hpp
```diff
@@ -369,13 +389,19 @@ struct HesaiCorrection : public HesaiCalibrationConfigurationBase
...
     unsigned int i = std::floor(1.f * azi / STEP3);
     unsigned int l = azi - i * STEP3;
     float k = 1.f * l / STEP3;
     return round((1 - k) * elevationOffset[ch * 180 + i] + k * elevationOffset[ch * 180 + i + 1]);
   }
+
+  [[nodiscard]] std::tuple<float, float> getFovPadding() const override
+  {
+    // TODO(mojomex): calculate instead of hard-coding
+    return {-5, 5};
```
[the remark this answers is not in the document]
mojomex: Addressed in [7272be2](https://github.com/tier4/nebula/pull/173/commits/7272be2936e9f381e66fc3b60b14c81fd3aa2b52)

nebula_tests/hesai/hesai_ros_scan_cutting_test_main.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+      if (has_points_near_end.find(p.channel) == has_points_near_end.end()) {
+        has_points_near_end.emplace(p.channel, false);
+      }
+      threshold_angle = drivers::normalize_angle(fov_max_rad - near_threshold_rad, 2 * M_PIf);
+      has_points_near_end[p.channel] |= drivers::angle_is_between(
+        threshold_angle, drivers::normalize_angle(fov_max_rad, 2 * M_PIf), p.azimuth);
+    }
+
+    bool is_at128 = drivers::SensorModelFromString(hesai_driver_->params_.sensor_model) ==
```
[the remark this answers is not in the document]
mojomex: Addressed in [eaa49f6](https://github.com/tier4/nebula/pull/173/commits/eaa49f697029f7a3306b2726782008a86073056a)

nebula_common/include/nebula_common/hesai/hesai_common.hpp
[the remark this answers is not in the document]
mojomex: Agreed :+1:

nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/angle_corrector_correction_based.hpp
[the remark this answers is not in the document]
mojomex: Addressed in [f4f8bf1](https://github.com/tier4/nebula/pull/173/commits/f4f8bf1f0c3f2a3abe7755641794d1f419c082c1)

mojomex:
@drwnz Thank you for the review!

I have addressed your comments and re-tested setting the `cut_angle` parameter to a value outside of the FoV.
Nebula now properly emits an error:
```
[INFO] [hesai_ros_wrapper_node]: OnParameterChange
[ERROR] [hesai_ros_wrapper_node]: Cannot cut scan outside of the FoV.
[WARN] [hesai_ros_wrapper_node]: OnParameterChange aborted: Could not set SensorConfiguration
```

The fix is in this commit: [2e42bd0](https://github.com/tier4/nebula/pull/173/commits/2e42bd045c3c2d5c3124ab7ea15b5ba4b8e94e6e)

nebula_hw_interfaces/include/nebula_hw_interfaces/nebula_hw_interfaces_hesai/hesai_hw_interface.hpp
```diff
@@ -331,6 +331,16 @@ class HesaiHwInterface
   /// @return Resulting status
   HesaiLidarRangeAll GetLidarRange();
 
+  /**
+   * @brief Given the HW interface's sensor configuration and a given calibration, set the sensor
+   * FoV (min and max angles) with appropriate padding around the FoV set in the configuration. This
+   * compensates for the points lost due to the sensor filtering FoV by raw encoder angle.
+   *
+   * @param calibration The calibration file of the sensor
+   * @return Status If the FoV was updated correctly
```
drwnz, line 340:
```suggestion
   * @return Status Resulting status of setting the FoV 
```

nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
```diff
@@ -532,6 +542,32 @@ HesaiLidarRangeAll HesaiHwInterface::GetLidarRange()
...
 }
 
+Status HesaiHwInterface::checkAndSetLidarRange(
+  const HesaiCalibrationConfigurationBase & calibration)
+{
+  if (sensor_configuration_->sensor_model == SensorModel::HESAI_PANDARAT128) {
+    return Status::SENSOR_CONFIG_ERROR;
+  }
+
+  int cloud_min = sensor_configuration_->cloud_min_angle * 10;
```
drwnz, line 552: Nit: Instead of multiplying by 10 and using 3600, it would be tidier to define decidegrees so it is clear when we are dealing in normal degrees or decidegrees.

nebula_ros/include/nebula_ros/hesai/hesai_ros_wrapper.hpp
```diff
@@ -29,17 +26,16 @@
 #include <rclcpp/rclcpp.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
 
-#include "nebula_msgs/msg/nebula_packet.hpp"
+#include <pandar_msgs/msg/pandar_scan.hpp>
```
drwnz, line 29: No need for nebula_packets and nebula_packet? (Ignore if not, I haven't gone through the include tree)

nebula_ros/include/nebula_ros/hesai/hesai_ros_wrapper.hpp
```diff
@@ -79,6 +78,23 @@ class HesaiRosWrapper final : public rclcpp::Node
   Status ValidateAndSetConfig(
     std::shared_ptr<const drivers::HesaiSensorConfiguration> & new_config);
 
+  /// @brief The ROS 2 parameter holding the calibration file path is called differently depending
+  /// on the sensor model. This function returns the correct parameter name given a model.
+  /// @param model A sensor model
```
drwnz, line 83:
```suggestion
  /// @param model The sensor model
```

nebula_ros/schema/sub/lidar_hesai.json
```diff
@@ -32,6 +32,20 @@
...
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 359,
+      "default": "0",
+      "description": "The angle in whole degrees which should be hit on top-of-second."
+    },
+    "cut_angle": {
+      "type": "number",
+      "minimum": 0,
+      "maximum": 360,
```
drwnz, line 45: [Future work] in the documentation it may be worth explaining why the cut angle can go to 360, but sync angle can not. Sync angle is an integer and cut angle is a float would be sufficient, it's just a little confusing how the sensor and driver handles 360 vs 0.

nebula_ros/src/hesai/decoder_wrapper.cpp
```diff
@@ -255,6 +131,12 @@ void HesaiDecoderWrapper::ProcessCloudPacket(
 
   cloud_watchdog_->update();
 
+  auto sys_pc_time_diff = parent_node_.now().seconds() - std::get<1>(pointcloud_ts);
+  // RCLCPP_INFO_STREAM(
```
drwnz, line 135: Leave commented code for easy debug later, or delete if no longer needed?

nebula_ros/src/hesai/hesai_ros_wrapper.cpp
```diff
@@ -88,9 +113,21 @@ nebula::Status HesaiRosWrapper::DeclareAndGetSensorConfigParams()
 
   {
     rcl_interfaces::msg::ParameterDescriptor descriptor = param_read_write();
-    descriptor.additional_constraints = "Angle where scans begin (degrees, [0.,360.])";
     descriptor.floating_point_range = float_range(0, 360, 0.01);
-    config.scan_phase = declare_parameter<double>("scan_phase", descriptor);
+    descriptor.description =
+      "At which angle to start a new scan. Cannot be equal to the start angle in a non-360 deg "
```
drwnz, line 118:
```suggestion
      "Angle at which to cut the scan. Cannot be equal to the start angle in a non-360 deg "
```

nebula_ros/src/hesai/hesai_ros_wrapper.cpp
```diff
@@ -292,6 +380,22 @@ rcl_interfaces::msg::SetParametersResult HesaiRosWrapper::OnParameterChange(
...
+    RCLCPP_INFO_STREAM(get_logger(), "Changed calibration to '" << new_cfg.calibration_path << "'");
+  }
+
+  if (
+    new_calibration_ptr && hw_interface_wrapper_ &&
+    sensor_cfg_ptr_->sensor_model != drivers::SensorModel::HESAI_PANDARAT128) {
+    auto status = hw_interface_wrapper_->HwInterface()->checkAndSetLidarRange(*new_calibration_ptr);
+    if (status != Status::OK) {
+      RCLCPP_ERROR_STREAM(
+        get_logger(), "Could not change FoV but config already changed: " << status);
```
drwnz, line 395: This error message is a little contradictory. Is it because everything has been set except range which failed?
In which case, something like "Sensor configuration updated, but FoV could not be set" might be better.

drwnz (commented):
I left a few more comments, nothing urgent.
I haven't gone through the tests except for verifying that they passed.

nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/angle_corrector_correction_based.hpp
```diff
@@ -53,9 +78,40 @@ class AngleCorrectorCorrectionBased : public AngleCorrector<HesaiCorrection>
     return field;
   }
 
+  uint32_t all_channels(uint32_t azi, double threshold, bool any, bool eq_ok)
```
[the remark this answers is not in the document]
mojomex: Addressed in [63233e7](https://github.com/tier4/nebula/pull/173/commits/63233e7686298ceb40fba031688d5d3bfbcad2a1)

nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/hesai_decoder.hpp
```diff
@@ -154,17 +172,38 @@ class HesaiDecoder : public HesaiScanDecoder
           }
         }
 
-        NebulaPoint point;
+        CorrectedAngleData corrected_angle_data =
+          angle_corrector_.getCorrectedAngleData(raw_azimuth, channel_id);
+        auto azimuth = corrected_angle_data.azimuth_rad;
```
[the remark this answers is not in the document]
mojomex: Addressed in 98808d4

nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/hesai_decoder.hpp
```diff
@@ -243,42 +273,48 @@ class HesaiDecoder : public HesaiScanDecoder
...
+        continue;
+      }
 
-      if (scan_completed) {
+      convertReturns(block_id, n_returns);
+
+      if (angle_corrector_.passedEmitAngle(last_azimuth_, block_azimuth)) {
         std::swap(decode_pc_, output_pc_);
-        decode_pc_->clear();
+        std::swap(decode_scan_timestamp_ns_, output_scan_timestamp_ns_);
```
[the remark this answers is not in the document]
mojomex: Added clarifications in 0bc2db9.
I agree the definitions are not strictly `decode` and `output` anymore (`deocde` and `overflow` from before also do not quite fit anymore).

However,

* the `decode` cloud is always used for decoding, except when cutting within the FoV and scans overlap
* in that case, the `output` and `decode` clouds are decoded to simultaneously in the overlap region
* the `output` cloud is always the one that is output when a scan is completed.

As I can't think of any better naming right now, I'd suggest leaving it as-is until we think of something better :bow:

nebula_ros/src/hesai/decoder_wrapper.cpp
mojomex: Removed in b54891e.

nebula_ros/src/hesai/hesai_ros_wrapper.cpp
mojomex: Re-worded in 7d4effa

nebula_ros/include/nebula_ros/hesai/hesai_ros_wrapper.hpp
mojomex: Good catch! Re-added in dbd8262

nebula_hw_interfaces/src/nebula_hesai_hw_interfaces/hesai_hw_interface.cpp
mojomex: Agreed. Since a lot of the values come from other places that are in degrees, the `* 10` is almost inevitable.
The cleanest (but for this PR overkill) approach would be proper types like `Degrees`, `DeciDegrees` which can be cast properly. See [std::chrono::duration](https://en.cppreference.com/w/cpp/chrono/duration) for nice unit handling.

drwnz (approved):
Ready to go, great work!

PR #178, opened 2024-07-22
fix: lower packet loss to near-zero levels

mojomex:
## PR Type

- Bug Fix

## Related Links

- [TIER IV internal link] -- internal ticket

## Description

This PR lowers the chance of packet loss occurring for all supported LiDARs by minimizing the number of mutex locks in the UDP receiver thread.

The UDP receiver thread was losing packets before, mainly because of scheduling:
1. Decoder is currently reading from queue
2. UDP receiver cannot acquire lock, its thread gets suspended
3. N UDP packets arrive (and are lost) while thread is not scheduled
4. When thread gets finally scheduled, it has lost packets

This is mostly mitigated by moving the `mt_queue` buffer from between the hardware interface and decoder to between the decoder and pointcloud/packets publishers. In other words, the queue is moved from the high-frequency part of Nebula (up to 36kHz for OT128) to the low-frequency part (10Hz or whatever the sensor framerate is set to).

The below figure shows the number of points in an unfiltered pointcloud (as a proxy for the number of packets received) for the baseline and fixed ("queue_only") implementations. The sensor tested was OT128 in dual return & high resolution mode.
The benchmarking setup was as follows (every run was repeated 3 times):
- taskset Nebula to cores 0-7
- taskset `ros2 bag record -s mcap --max-cache-size 5000000000 /pandar_points` to cores 8-11
- run 8 instances of `ros2 topic echo  /pandar_points --field header` taskset to core `12 + 2i` for instance i
- stop all processes after 60s

The plot was generated from the concatenated rosbags.

nebula_queue

|  | messages | mean | std | min | 0.1% | 1% | 10% | 50% |
|-|-:|-:|-:|-:|-:|-:|-:|-:|
| baseline |1715 |920577 |2436.64 |865024 |895010 |911908 |919040 |**921600** |
| fix |1717|921585 |582.054 | 897536| 920503|**921600** |**921600** | **921600** |
| baseline (% lost) |  |920577 |0.11  |6.14 |2.89 |1.05 |0.28 |**0** |
| fix (% lost) | |0.00 | | 2.61| 0.12|**0** |**0** | **0** |

:information_source: This PR applies to all Hesai, Velodyne, and Robosense sensors.
:information_source: #169 (Aeva Aeries II) already uses a similar approach and should be immune to packet loss
:warning: I am not sure at the moment whether this change would benefit Continental sensors as well (HW monitor and other functions handled in decoder, packet rate unknown to me), so I skipped them for now. @knzo25 have you observed packet loss on Radars so far?

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/hesai_packet.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/hesai_packet.hpp
@@ -17,6 +17,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <ctime>
+#include <stdexcept>
 
 namespace nebula
 {
--- a/nebula_decoders/src/nebula_decoders_hesai/hesai_driver.cpp
+++ b/nebula_decoders/src/nebula_decoders_hesai/hesai_driver.cpp
@@ -13,6 +13,8 @@
 #include "nebula_decoders/nebula_decoders_hesai/decoders/pandar_xt32.hpp"
 #include "nebula_decoders/nebula_decoders_hesai/decoders/pandar_xt32m.hpp"
 
+#include <rclcpp/logging.hpp>
+
 // #define WITH_DEBUG_STD_COUT_HESAI_CLIENT // Use std::cout messages for debugging
 
 namespace nebula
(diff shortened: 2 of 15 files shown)


mojomex:
This was not needed after all and the same can be achieved by allocating larger network buffers (`net.core.rmem_default`).

PR #182, opened 2024-08-13
feat(hesai): add a ring section filter to Hesai sensors

mojomex:
## PR Type

- New Feature

## Related Links

- [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-7133) -- related ticket
- #69 -- similar PR for Velodyne

## Description

This PR introduces a simple and extendable framework to implement per-point filters in Nebula. The first filter, also added in this PR, is a ring section filter.

### Point Filters

A point filter has one member function `bool excluded(const NebulaPoint &)` which is called per point, right after that point has been fully decoded (including angle correction). The point is excluded from the pointcloud as soon as any of the configured filters excludes it. Filters run sequentially and later filters are not executed if an earlier filter has exluded the point already.

The filter pipeline is configured via JSON, in a format like `{"filter1": filter_params1, "filter2": filter_params2}` and each filter type can only appear once.

### Ring Section Filter

The filter can be configured to cut out zero or more sections of each ring, each section having a start and end angle.
The configuration looks like this: `[[channel_id, start_deg, end_deg], [channel_id, start_deg, end_deg], ...]`. Not all channels have to be specified, and channels can be specified multiple times. The start and end values are floating-point, so precise angles can be defined. Sections can overlap, but no optimizations are done on Nebula's side to merge overlapping regions.

Angles greater than 360deg and smaller than 0deg can be specified and will be auto-normalized by Nebula. Sections spanning over the 360/0 boundary are also handled correctly.

### Performance

This is a fairly naive implementation:
* while parsing, an array of arrays of sections is created. The index in the outer array corresponds to the channel ID, while the inner arrays are a collection of their rings' sections.
* for each point, all sections on its ring are checked until it is excluded or all sections have been checked.

:construction: Performance plots follow

- [ ] Assign PR to reviewer

## CI Checks


mojomex:
Closing in favor of #251.

PR #186, opened 2024-08-27
feat(hesai_hw_interface): configure and check `SO_RCVBUF` automatically

mojomex:
## PR Type

- Improvement

## Description

High-bandwidth sensors such as the OT128 may experience packet loss due to the UDP socket's receive buffer being too small by default. This PR adds functionality to set and confirm that buffer size.

The buffer has to be set to a size that
1. is as small as possible to save memory
2. large enough to store all packets that arrive while the UDP receiver thread is not scheduled or has its callback running

For now, the chosen (hard-coded) size is `1500 B * 3600 = 5.4 MB`, which is the MTU size (`1500 B`) times the maximum number of packets in one full scan of OT128.
The memory footprint is still 'kind of' small compared to the size of pointcloud buffers etc., so condition 1 is fulfilled for the moment. In the future, computing the optimal buffer size based on sensor bandwidth would be preferable.

## Review Procedure

Setting `rmem_default` and `rmem_max` is done like: `sudo sysctl -w net.core.rmem_default=123`.

* Set `rmem_default` and `rmem_max` to a size smaller than `5.4 MB`, e.g. `500000` for `500 KB`. Nebula should refuse to launch with an error message telling you to increase `rmem_max`.
* Set `rmem_default` to a small size like above, and set `rmem_max` to a value greater than `5.4 MB`, e.g. `1000000000` for `1 GB`. Nebula should now run, set its socket's buffer size to `5.4 MB` and there should be no packet loss.

- [x] Assign PR to reviewer

## CI Checks


nebula_ros/include/nebula_ros/continental/continental_ars548_hw_interface_wrapper.hpp
```diff
@@ -26,6 +26,7 @@
 #include <continental_srvs/srv/continental_ars548_set_vehicle_parameters.hpp>
 #include <geometry_msgs/msg/accel_with_covariance_stamped.hpp>
 #include <geometry_msgs/msg/twist_with_covariance_stamped.hpp>
+#include <std_msgs/msg/float32.hpp>
```
knzo25, line 29: Why is this needed here?

nebula_hw_interfaces/include/nebula_hw_interfaces/nebula_hw_interfaces_hesai/hesai_cmd_response.hpp
```diff
@@ -21,6 +21,7 @@
 
 #include <array>
 #include <cstdint>
+#include <iomanip>
```
knzo25, line 24: Why is this needed?

knzo25 (changes requested):
Left some small questions :pray:

nebula_ros/include/nebula_ros/continental/continental_ars548_hw_interface_wrapper.hpp
mojomex: Float32 is used in line 69. Maybe some other included header had previously been importing it, but due to some unrelated change, this became necessary here.

nebula_hw_interfaces/include/nebula_hw_interfaces/nebula_hw_interfaces_hesai/hesai_cmd_response.hpp
mojomex: `std::setw` and `std::setfill` further down

nebula_ros/include/nebula_ros/continental/continental_ars548_hw_interface_wrapper.hpp
knzo25: Mmm did not get the answer, but the diff became outdated so I will mark this as solved

knzo25 (approved):
LGTM

knzo25:
@mojomex

PR #193, opened 2024-09-05
fix(at128): allow negative values in points' `elevation` fields

mojomex:
## PR Type

- Improvement

## Related Links

- #173 -- the base for this PR, needed because AT128 is broken before this PR

## Description

:warning: This has to be rebased onto develop after #173 is merged, and then be merged.

Only AT128 normalized its points' `elevation` field to [0, 2pi) while other sensors allowed negative `elevation` values. While the normalization is done on azimuth as well and thus, for consistency sake, would be nice on the elevation field, too, this causes visualizations to look ugly when coloring by `elevation`. 
Anyway, trigonometry functions won't care about this change as they are periodic, and given that all other sensors are not normalizing elevation, it is safe to assume this doesn't impact downstream modules.

| Old (normalized) | New (unnormalized) |
|-|-|
| image | image |

## Review Procedure

Confirm the `elevation` field looks correct. Code review.

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/angle_corrector_correction_based.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/angle_corrector_correction_based.hpp
@@ -181,19 +181,23 @@ public:
   {
     int field = findField(block_azimuth);
 
-    int32_t elevation =
-      correction_->elevation[channel_id] +
-      correction_->getElevationAdjustV3(channel_id, block_azimuth) * (AngleUnit / 100);
+    int32_t elevation = correction_->elevation[channel_id] +
+                        correction_->getElevationAdjustV3(channel_id, block_azimuth) *
+                          static_cast<int32_t>(AngleUnit / 100);
+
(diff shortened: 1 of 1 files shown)


knzo25:
@mojomex 
https://github.com/tier4/nebula/pull/173 was merged, so could you rebase or solve conflicts before checking this PR?

mojomex:
@knzo25 Thank you for the reminder, done. :+1:

mojomex:
@knzo25 As discussed internally, since this is a small 'hotfix'-style change, this is fine to merge into main.
I have tested the PR with a live sensor during the `develop-eval` tests and confirmed it working (see screenshots above). Code review should be enough imho.

knzo25 (approved):
LGTM

question though:
You are using `MAX_AZIMUTH` to normalize `elevation`.
I know that it is set to `360 * AngleUnit` but if there comes a time where that stops being true (separate directly solid state from mechanicals, etc), it would be worth considering using a separate variable

mojomex:
@knzo Thanks for the input! The angle corrector classes are very sensor-specific (e.g. only AT128) so I think we are fine for some time to come. Other sensors such as the Aeries II don't have such variables at all for example.

PR #194, opened 2024-09-10
docs: create API reference page, link class/namespace/... references in navigation

mojomex:
## PR Type

- Improvement

## Description

This PR should make it much user-friendlier to navigate through the API reference.

image

## Review Procedure

Run
```shell
python3 -m pip install -U $(curl -fsSL https://raw.githubusercontent.com/autowarefoundation/autoware-github-actions/main/deploy-docs/mkdocs-requirements.txt)
cd src
mkdocs serve
```
and confirm the API reference tab is working as expected.

## Pre-Review Checklist for the PR Author

**PR Author should check the checkboxes below when creating the PR.**

- [x] Assign PR to reviewer

## Checklist for the PR Reviewer

**Reviewers should check the checkboxes below before approval.**

- [ ] Commits are properly organized and messages are according to the guideline
- [ ] (Optional) Unit tests have been written for new behavior
- [ ] PR title describes the changes

## Post-Review Checklist for the PR Author

**PR Author should check the checkboxes below before merging.**

- [ ] All open points are addressed and tracked via issues or tickets

## CI Checks

- **Build and test for PR**: Required to pass before the merge.

PR #195, opened 2024-09-11
docs: add parameter tables generated from schemas per vendor and sensor model

mojomex:
## PR Type

- Improvement

## Description

:warning: Merge #194 first! :warning: 

This PR adds 
* a modified version of Autoware.Universe's [mkdocs_macros.py](https://github.com/autowarefoundation/autoware.universe/blob/main/mkdocs_macros.py) to add support for Markdown table generation from JSON schemas
* a page per vendor with common parameters
* a page per sensor with all the parameters (and value ranges) supported by that sensor

image

## Review Procedure

Render the docs and confirm the tables are complete and correct.

- [x] Assign PR to reviewer

## CI Checks


drwnz (changes requested):
Mostly looks great, some minor suggested changes.
The only other thing that I noticed, was that generally parameter tables are built from the schema except for Hesai common which has extra (very detailed) explanation.
It might be more consistent to generate Hesai common the same as the others and add an explanation below the parameter table. This could also be done for the Nebula common parameters as well.

mojomex:
The Hesai table I can do.

For the common one, there is almost no overlap between **all** sensors, so I would have to create around 3 separate schemas to cover everything (e..g CAN vs. IP, scan_phase vs cut angle etc.). I don't think it would be a particularly nice solution.

mojomex:
I added the Hesai table, please check @drwnz :pray: 

image

drwnz (approved):
LGTM, nice :)

PR #204, opened 2024-10-01
chore: fix typos in the SetVelocityVehicle function

JonasKrug:
## PR Type

- Bug Fix

## Description

It fixes the naming of an internal variable in the continental ars548 hardware interface code.

## Review Procedure

Check that all occurrences of the misnamed variable were properly replaced.


knzo25 (approved):
LGTM !
Thanks for the catch

PR #205, opened 2024-10-02
chore: fix case styles

mojomex:
## PR Type

- Improvement

## Related Links

- [the Clang-Tidy config](https://github.com/autowarefoundation/autoware/blob/main/.clang-tidy)

## Description

Case styles in Nebula are not consistent, and Clang-Tidy marks almost every file as having warnings because of that.

This will be important for enforcing coding styles in contributed code as well.

:green_circle: This PR contains semi-automatic changes only and has been confirmed to build. Thus, a cursory review is enough.

## Review Procedure

No special review necessary.

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,7 +87,9 @@ repos:
     rev: 1.6.1
     hooks:
       - id: cpplint
-        args: [--quiet]
+        # runtime/arrays uses the name of the variable to determine const-ness.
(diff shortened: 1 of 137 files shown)


drwnz (approved):
Confirmed that it builds and runs without issue. It's a bit scary to see everything become snake case but seems that's what ROS orders, so I guess it's OK!

Issue #136, opened 2024-04-08
`setup_sensor` flag doesn't disable sensor configuration communication


msz-rai:
According to the README ([Generic launch file](https://github.com/tier4/nebula?tab=readme-ov-file#generic-launch-file) section), `setup_sensor` flag should disable sensor configuration communication. It doesn't work in my case. I am running the following command:
```
ros2 launch nebula_ros nebula_launch.py sensor_model:=PandarQT64 setup_sensor:=false sensor_ip:=127.0.0.1 return_mode:=First
```
and `hesai_hw_interface_ros_wrapper_node` is trying to connect with the sensor until successful. I get the following output:

```
[component_container-1] [INFO] [1712565931.610202378] [hesai_hw_interface_ros_wrapper_node]: SensorConfig:SensorModel: PandarQT64, FrameID: hesai, HostIP: 255.255.255.255, SensorIP: 127.0.0.1, DataPort: 2368, ReturnMode: First, Frequency: 0, MTU: 1500, Use sensor time: 0, GnssPort: 10110, ScanPhase:0, RotationSpeed:600, FOV(Start):0, FOV(End):360, DualReturnDistanceThreshold:0.1, PtpProfile:IEEE_1588v2, PtpDomain:0, PtpTransportType:UDP/IP, PtpSwitchType:TSN
[component_container-1] 127.0.0.1:9347
[component_container-1] [ERROR] [1712565931.610316207] [TcpSocket::open]: Connection refused
[component_container-1] [INFO] [1712565931.610333148] [hesai_hw_interface_ros_wrapper_node]: hesai_hw_interface_ros_wrapper_node Retry: 0
[component_container-1] [ERROR] [1712565939.610506246] [hesai_hw_interface_ros_wrapper_node]: hesai_hw_interface_ros_wrapper_node Retry: 1
[component_container-1] [ERROR] [1712565939.610584644] [TcpSocket::shutdown]: Bad file descriptor
[component_container-1] 127.0.0.1:9347
[component_container-1] [ERROR] [1712565939.610664718] [TcpSocket::open]: Connection refused
[component_container-1] [ERROR] [1712565947.610774567] [hesai_hw_interface_ros_wrapper_node]: hesai_hw_interface_ros_wrapper_node Retry: 2
[component_container-1] [ERROR] [1712565947.610876308] [TcpSocket::shutdown]: Bad file descriptor
[component_container-1] [ERROR] [1712565947.610963187] [TcpSocket::open]: Connection refused
[...]

PR #210, opened 2024-10-16
feat: allow to launch hardware in listen mode

msz-rai:
## PR Type

- Improvement

## Related Links

https://github.com/tier4/nebula/issues/136

## Description

This PR introduces a new parameter to the `hesai_ros_wrapper` and `velodyne_ros_wrapper` nodes, allowing hardware to launch with communication to the sensor disabled. In this configuration, the driver operates in a "listen-only" mode, receiving packet data without sending commands.

This functionality is required for the [AWSIM](https://github.com/tier4/AWSIM) simulator, where the LiDAR simulation plugin supports publishing raw packets via UDP but doesn't handle communication from the driver.

Prior to nebula version `0.2.0`, there was a workaround for this behaviour (as outlined in the related issue). However, recent architectural changes have removed the ability to use that workaround (segmentation fault appears).

## Review Procedure

Code review + test to ensure that the new parameter does not interfere with the normal operation of real-world sensors.

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/src/hesai/hw_interface_wrapper.cpp
+++ b/nebula_ros/src/hesai/hw_interface_wrapper.cpp
@@ -9,10 +9,11 @@ namespace nebula::ros
 
 HesaiHwInterfaceWrapper::HesaiHwInterfaceWrapper(
   rclcpp::Node * const parent_node,
-  std::shared_ptr<const nebula::drivers::HesaiSensorConfiguration> & config)
+  std::shared_ptr<const nebula::drivers::HesaiSensorConfiguration> & config, bool communicate_with_sensor)
 : hw_interface_(new nebula::drivers::HesaiHwInterface()),
   logger_(parent_node->get_logger().get_child("HwInterface")),
-  status_(Status::NOT_INITIALIZED)
+  status_(Status::NOT_INITIALIZED),
+  communicate_with_sensor_(communicate_with_sensor)
 {
   setup_sensor_ = parent_node->declare_parameter<bool>("setup_sensor", param_read_only());
   bool retry_connect = parent_node->declare_parameter<bool>("retry_hw", param_read_only());
@@ -27,6 +28,11 @@ HesaiHwInterfaceWrapper::HesaiHwInterfaceWrapper(
 
   hw_interface_->SetLogger(std::make_shared<rclcpp::Logger>(parent_node->get_logger()));
   hw_interface_->SetTargetModel(config->sensor_model);
+
+  if (!communicate_with_sensor_) {
+    // No need to initialize Tcp if communication is disabled
+    return;
+  }
 
   int retry_count = 0;
 
@@ -63,6 +69,10 @@ HesaiHwInterfaceWrapper::HesaiHwInterfaceWrapper(
 void HesaiHwInterfaceWrapper::on_config_change(
   const std::shared_ptr<const nebula::drivers::HesaiSensorConfiguration> & new_config)
 {
+  if (!communicate_with_sensor_) {
+    RCLCPP_ERROR_STREAM(logger_, "Cannot change sensor configuration: communication with sensor is disabled");
+    return;
+  }
   hw_interface_->SetSensorConfiguration(
     std::static_pointer_cast<const nebula::drivers::SensorConfigurationBase>(new_config));
   if (setup_sensor_) {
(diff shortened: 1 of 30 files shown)


nebula_ros/src/velodyne/hw_interface_wrapper.cpp
```diff
@@ -23,6 +25,11 @@ VelodyneHwInterfaceWrapper::VelodyneHwInterfaceWrapper(
       (std::stringstream{} << "Could not initialize HW interface: " << status_).str());
   }
 
+  if (!communicate_with_sensor_) {
+    // Do not initialize http client if communication is disabled
+    return;
```
mojomex, line 30: The `status_` is not set before returning, resulting in it being `NOT_INITIALIZED` instead of the expected `OK`.

nebula_ros/src/hesai/hw_interface_wrapper.cpp
```diff
@@ -28,6 +30,11 @@ HesaiHwInterfaceWrapper::HesaiHwInterfaceWrapper(
   hw_interface_->SetLogger(std::make_shared<rclcpp::Logger>(parent_node->get_logger()));
   hw_interface_->SetTargetModel(config->sensor_model);
 
+  if (!communicate_with_sensor_) {
+    // Do not initialize Tcp if communication is disabled
+    return;
```
mojomex, line 35: `status_` is not updated before returning. Should be `Status::OK`.

mojomex (changes requested):
Thank you for your great work!

The [json-schema-check](https://github.com/tier4/nebula/actions/runs/11371025364/job/31632149628?pr=210) is currently failing, could you add the new parameter ton the affected sensor model and vendor schemas in `nebula_ros/schema`?

As for the parameter name, I think that `communicate_with_sensor` can easily be confused with `launch_hw`, as in, users could think that this affects ALL communication including UDP.
Imho, a more fitting name would be `disable_tcp` or `udp_only`.

Please also add a info-level log message in the wrappers that is printed when TCP is disabled, noting that checking and synchronizing settings, and diagnostics publishing are disabled.

Otherwise everything looks good to go!

nebula_ros/src/velodyne/hw_interface_wrapper.cpp
msz-rai: The `status_` should be set after sensor configuration is initialized ([here](https://github.com/msz-rai/nebula/blob/feature/disable-communication-option/nebula_ros/src/velodyne/hw_interface_wrapper.cpp#L20)).

nebula_ros/src/hesai/hw_interface_wrapper.cpp
msz-rai: The `status_` should be set after sensor configuration is initialized ([here](https://github.com/msz-rai/nebula/blob/feature/disable-communication-option/nebula_ros/src/hesai/hw_interface_wrapper.cpp#L21)).

msz-rai:
@mojomex 

Thank you for the review!

I agree that my parameter name suggestion was misleading. I chose `udp_only` instead of `disable_tcp` because Velodyne uses HTTP for communication.

I also applied the rest of review suggestions. PR is ready for re-review.

mojomex:
@msz-rai Thanks for your quick response,
I'll re-review it tomorrow!

nebula_ros/src/hesai/hw_interface_wrapper.cpp
mojomex: Ah, indeed. My bad!

nebula_ros/src/velodyne/hw_interface_wrapper.cpp
mojomex: Ah, indeed. My bad!

mojomex (commented):
LGTM!

Before we merge, we had an internal discussion because there's now three connection-related parameters (`udp_only`, `launch_hw` and `setup_sensor`). Having three booleans (= 8 possible combinations) for only four valid parameter combinations (`launch_hw && setup_sensor && !udp_only`, `launch_hw && !setup_sensor && !udp_only`, `launch_hw && !setup_sensor && udp_only`, `!launch_hw`) will probably be confusing for users.

We'd like to replace the three parameters with one parameter, `connection_mode`, which is one of `offline`, `udp_only`, `skip_setup`, `full`, for clarity.

Since this is out of scope for this PR, I'll make that change in another PR immediately after this one is merged. @msz-rai do you have any objections to this?

msz-rai:
Everything looks good to me!

mojomex (dismissed):
LGTM!

mojomex (commented):
I'll test with a sensor tomorrow and then merge :bow:

mojomex (approved):
:+1:

PR #239, opened 2024-11-27
fix(hesai): set PTP lock offset on sensor configuration

mojomex:
## PR Type

- Improvement
- Bug Fix

## Related Links

- [TIER IV INTERNAL LINK](https://star4.slack.com/archives/C076E8EBJTG/p1731645115682099) - corresponding sensing inquiry

## Description

This PR adds support for the `SET_PTP_LOCK_OFFSET` and `GET_PTP_LOCK_OFFSET` PTC commands.
In the inquiry linked above, it was observed that not explicitly setting PTP lock offset sometimes causes the value to reset to `1 us` on at least OT128. While this is likely a sensor firmware bug, being able to set this offset from within Nebula should improve usability, and should eliminate the need to use Hesai's web interface or desktop GUI.

- [ ] Assign PR to reviewer

## CI Checks


mojomex:
# Evaluation

| Sensor Model | Works? |
|-|:-:|
| OT128 | :green_circle:  |

Here is a demo of OT128:
https://github.com/user-attachments/assets/4145746e-ca97-48f9-a66c-cf98b7d5ba91

It has been confirmed that by restarting Nebula, for a brief moment, the setting can revert to the old value (Hesai firmware bug or PTC documentation inaccurate). With this PR, it is immediately set to the desired value again.

mojomex:
Restricting this feature to only affect OT128 and QT128, as other sensors have been found to not support it (maybe a firmware update would change that).

Sadly, the current architecture for sensor configurations in Nebula forces us to add the parameter to ALL Hesai sensors, even if without any effect.

drwnz (changes requested):
Please add to the documentation as well, but otherwise LGTM!

TomohitoAndo (approved):
> Please add to the documentation as well, but otherwise LGTM!

If this is done, LGTM.

Issue #243, opened 2024-12-03
LiDAR info topic


drwnz:
In a ROSBag with only `nebula_packets` there is no way to know the LiDAR configuration that generated these packets.
Especially settings which change the pointcloud, like dual return settings, have to be known through also supplying the launchfile used with any ROSBag, for example.

Additionally, device-specific LiDAR calibration files are downloaded from the sensor or pre-downlaoded files read from disk, but the user has to manage the calibration files. Again, portability becomes an issue.

Therefore, we would like to propose a LiDAR info topic (something like the camera info topic) that periodically publishes the LiDAR calibration and configuration information. This requires adding a new message type as well. A static topic could also be considered but modifying and cutting ROSbags needs extra handling.

PR #244, opened 2024-12-03
fix(hesai_ros_wrapper): demote calibration fallback message to WARN

mojomex:
## PR Type

- Improvement

## Related Links

- [TIER IV INTERNAL LINK](https://star4.slack.com/archives/C076E8EBJTG/p1733099784725949) -- internal discussion
- #243 -- this feature would rid us of the need to manage calibration files completely

## Description

When Nebula runs without a live Hesai sensor, it will search for an already-downloaded calibration file. If that is not found, it will log a message which was previously at log level `ERROR`.
As quite a common workflow is to decode recorded packets on a different machine without this downloaded file, it is more reasonable to demote this to `WARN`.

After the warning is emitted, Nebula falls back to a sensor-specific, generic calibration file and continues operation.

With #243, calibration data could be recorded alongside `/pandar_packets` and replayed on any machine, making this file handling routine obsolete. Pointclouds could also be perfectly reproduced without any extra steps.

- [ ] Assign PR to reviewer

## CI Checks


drwnz (approved):
LGTM!

PR #250, opened 2025-01-15
feat: mask-based pruning filter (vendor independent)

mojomex:
## PR Type

- New Feature

## Related Links

- Better version of #182

## Description

:information_source: This PR provides only a library, and does not have an effect by itself. Vendor implementations are done in PRs following this one.

:information_source: Performance measurements are provided on a per-vendor basis (e.g. #251) and are thus not included in this PR.

This PR introduces a mask-based pointcloud filter that allows precise pruning of the pointcloud with different pruning factors for different areas of the pointcloud.

This feature is aimed at reducing computational load later in the pipeline by
* offering a replacement to filters like CropBox
* providing a way to downsample pointcloud angular resolution in certain areas

The filter takes a grayscale PNG image with the x direction representing azimuth,  and the y direction representing the channel - a representation for elevation instead of channel will be added once such sensors are in Nebula.

Points falling into 
* white regions (`255 = 100 %`) are all kept
* black regions (`0 = 0 %`) are all discarded
* gray regions are filtered according to a dithered black-and-white version of the mask computed on startup

The dithering algorithm first quantizes the grayscale image into multiples of `1/10`, yielding `11` quantization levels.
Dithering patterns are diagonal lines spaced as uniformally as possible in a 10-pixel period (both horizontally and vertically).

See the [docs](https://github.com/tier4/nebula/blob/5b487be03ca4e1be50b03e20dc76c29676893e44/docs/filters.md) for images and behavior descriptions.

## Review Procedure

* code review
* verify what is being verified with unit tests
* run the filter in one of the upcoming vendor implementations (e.g. #251)

- [x] Assign PR to reviewer

## CI Checks


nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
```diff
@@ -0,0 +1,143 @@
...
+}  // namespace impl
+
+class DownsampleMaskFilter
+{
+  static const uint8_t g_quantization_levels = 10;
+
+public:
+  DownsampleMaskFilter(
+    const std::string & filename, AngleRange<int32_t> azimuth_range_mdeg,
+    uint32_t azimuth_peak_resolution_mdeg, size_t n_channels,
```
drwnz, line 91: Might be best to just in case add a check that this value is not zero, as it is user-supplied.

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
```diff
@@ -0,0 +1,143 @@
...
+    const std::shared_ptr<loggers::Logger> & logger, bool export_dithered_mask = false)
+  : azimuth_range_{deg2rad(azimuth_range_mdeg.min / 1000.), deg2rad(azimuth_range_mdeg.max / 1000.)}
+  {
+    png::image<png::gray_pixel> factors(filename);
+
+    size_t mask_cols = azimuth_range_mdeg.extent() / azimuth_peak_resolution_mdeg;
```
drwnz, line 97: There's nothing stopping extent() from returning a negative number when max < min (see previous comment) might give something unexpected for size_t.

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
```diff
@@ -0,0 +1,143 @@
...
+  static const uint8_t g_quantization_levels = 10;
+
+public:
+  DownsampleMaskFilter(
+    const std::string & filename, AngleRange<int32_t> azimuth_range_mdeg,
+    uint32_t azimuth_peak_resolution_mdeg, size_t n_channels,
+    const std::shared_ptr<loggers::Logger> & logger, bool export_dithered_mask = false)
+  : azimuth_range_{deg2rad(azimuth_range_mdeg.min / 1000.), deg2rad(azimuth_range_mdeg.max / 1000.)}
+  {
+    png::image<png::gray_pixel> factors(filename);
```
drwnz, line 95: Might want to check that the file could be read here - or does the png library do a good job of this? Either way, no point in continuing if factors doesn't exist. Although it will get caught by the size check I imagine.

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
```diff
@@ -0,0 +1,143 @@
...
+        logger->info("Wrote dithered mask to " + out_path.native());
+      } catch (const png::std_error & e) {
+        logger->warn("Could not write " + out_path.native() + ": " + e.what());
+      }
+    }
+  }
+
+  bool excluded(const NebulaPoint & point)
+  {
+    double azi_normalized = (point.azimuth - azimuth_range_.min) / azimuth_range_.extent();
```
drwnz, line 127: extent() can potentially return zero (if the user is silly)

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+    auto x = static_cast<ssize_t>(std::round(azi_normalized * static_cast<double>(mask_.cols())));
```
drwnz, line 129: Might be missing something, but I don't see why a <ssize_t> is used here when a <size_t> and any required clamping would be safer (as x should never be negative).

drwnz (commented):
I added some comments - mostly around the use of the added AngleRange. 
Testing looks good!

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
mojomex: The png++ library will throw a `png::std_error`. Currently, the check if the file exists is done in the Hesai wrapper but because this is a constructor, there's not much else we can do here.

Catching the `std_error` when instantiating the filter would be good though, will implement in the other PR #251

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
mojomex: Addressed in 8c4e0d3

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
mojomex: Addressed in 8c4e0d3

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
mojomex: Addressed in 8c4e0d3

nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
mojomex: `azi_normalized` can be smaller than 0, making the rounded result also smaller than 0.
I personally find it cleaner to not rely on the negative value wrapping around to a huge positive one.
I am checking for negativity in `x_out_of_bounds` before I access the mask, so this is safe even though `x` is a `ssize_t`.

nebula_decoders/tests/point_filters/test_downsample_mask.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <filesystem>
+#include <limits>
+#include <memory>
+#include <string>
+
```
drwnz, line 33: Pre-commit is also requesting 
`#include <vector>`

mojomex: Addressed in a85ea4b

nebula_decoders/include/nebula_decoders/nebula_decoders_common/angles.hpp
```diff
@@ -18,6 +18,29 @@
...
+  T elevation;
+};
+
+template <typename T>
+struct AngleRange
+{
+  T min;
+  T max;
+
+  [[nodiscard]] T extent() const { return max - min; }
```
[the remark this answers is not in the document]
mojomex: Addressed in cec2c76

mojomex:
@drwnz I've addressed the rest of your feedback, please have a look :smiley:

nebula_decoders/include/nebula_decoders/nebula_decoders_common/angles.hpp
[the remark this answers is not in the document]
drwnz: Nice :+1:

drwnz (approved):

PR #279, opened 2025-02-20
fix(velodyne): remove heavy get_snapshot from HW monitor

mojomex:
## PR Type

- Bug Fix

## Related Links

- [TIER IV INTERNAL](https://star4.slack.com/archives/C076E8EBJTG/p1739770915233339) -- the reported issue

## Description

The `/cgi/snapshot.hdl` endpoint of the sensor is not meant to be polled every second. Doing that causes the sensor to sometimes drop communication for a while (see the linked issue above).

This hotfix removes the timer polling `get_snapshot` and the related, now unused code.
As far as I can tell, marking diag info stale did not work reliably before (updates from `get_diag` and `get_status` were ignored, and will not work at all now.
All diag information will still be output as before.

## Review Procedure

Test with a real VLP16, VLP32 and VLS128 and compare before/after, especially with advanced diag enabled.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/src/velodyne/hw_monitor_wrapper.cpp
+++ b/nebula_ros/src/velodyne/hw_monitor_wrapper.cpp
@@ -30,18 +30,19 @@ VelodyneHwMonitorWrapper::VelodyneHwMonitorWrapper(
   std::cout << "Get model name and serial." << std::endl;
   auto str = hw_interface_->get_snapshot();
   if (!str.has_value()) return;
-  current_snapshot_time_.reset(new rclcpp::Time(parent_node_->now()));
-  current_snapshot_tree_ =
+
+  auto snapshot_tree =
     std::make_shared<boost::property_tree::ptree>(hw_interface_->parse_json(str.value()));
   current_diag_tree_ =
-    std::make_shared<boost::property_tree::ptree>(current_snapshot_tree_->get_child("diag"));
+    std::make_shared<boost::property_tree::ptree>(snapshot_tree->get_child("diag"));
   current_status_tree_ =
-    std::make_shared<boost::property_tree::ptree>(current_snapshot_tree_->get_child("status"));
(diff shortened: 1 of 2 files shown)


nebula_ros/src/velodyne/hw_monitor_wrapper.cpp
```diff
@@ -164,55 +161,8 @@ void VelodyneHwMonitorWrapper::initialize_velodyne_diagnostics()
...
-
-    current_snapshot_time_.reset(new rclcpp::Time(parent_node_->now()));
-    current_snapshot_tree_ = std::make_shared<boost::property_tree::ptree>(ptree);
-    current_diag_tree_ =
-      std::make_shared<boost::property_tree::ptree>(current_snapshot_tree_->get_child("diag"));
-    current_status_tree_ =
-      std::make_shared<boost::property_tree::ptree>(current_snapshot_tree_->get_child("status"));
-    current_snapshot_.reset(new std::string(str.value()));
-  }
+  diagnostics_update_timer_ = parent_node_->create_wall_timer(
```
interimadd, line 164: This line can be replaced with the following method:

`diagnostics_updater_.setPeriod(1.0);`

This simplifies the code. However, I don't believe this change is mandatory, so I'll approve it as is.

interimadd (approved):
Thank you for fixing issue for diagnostics. 
I've added a single comment for a small suggestion, but totally looks great so I approved this  PR.

KYabuuchi (approved):
Thanks for creating the fix PR!
I tested this fix using both the VLP-16 and VLS-128, and it looks good to me.

nebula_ros/src/velodyne/hw_monitor_wrapper.cpp
mojomex: Thank you, addressed in 4a385f8!

PR #251, opened 2025-01-16
feat(hesai): implement mask-based pruning filter for Hesai

mojomex:
## PR Type

- New Feature

## Related Links

- #250 -- based on this PR

## Description

This PR implements the mask-based pruning filter introduced in #250 for all Hesai sensors. For more specifics on sensor configuration changes and filter behavior, see #250.

## Review Procedure

* Test whether pointcloud output is as expected when not specifying a filter
* Try and create a downsample mask, e.g. in GIMP, and see if it works expected

## Pre-Review Checklist for the PR Author

**PR Author should check the checkboxes below when creating the PR.**

- [ ] Assign PR to reviewer

## Checklist for the PR Reviewer

**Reviewers should check the checkboxes below before approval.**

- [ ] Commits are properly organized and messages are according to the guideline
- [ ] (Optional) Unit tests have been written for new behavior
- [ ] PR title describes the changes

## Post-Review Checklist for the PR Author

**PR Author should check the checkboxes below before merging.**

- [ ] All open points are addressed and tracked via issues or tickets

## CI Checks

- **Build and test for PR**: Required to pass before the merge.

PR #287, opened 2025-03-07
fix(ot128): introduce anti-flicker downsample dithering pattern

mojomex:
## PR Type

- Improvement

## Related Links

- #251 -- introduced mask based pruning filter

## Description

:warning: The below use cases are **not** affected by the flickering issue (already working fine):
* masks converted from #182
* black-and-white masks
* OT128 in standard/non-high-resolution mode
* other sensors

While the mask based pruning filter introduced in #251 works well for all supported Hesai sensors, including OT128 in standard mode, OT128's high resolution mode employs a dithering pattern that interferes with the pruning filter's one.

This results in flicker and could degrade perception performance etc.
This PR extends the `DownsampleMaskFilter` by allowing sensors to specify a transform function for azimuth/channel coordinates passed to the dithering algorithm. This allows the developer to include knowledge about the sensor's dithering/scan pattern into the mask dithering step.
All sensors except OT128 are using the default transform, as before, with only OT128 using a specialized one.

For details on how OT128 dithers, see the chapter _B.4. Laser firing time of each channel_ of the [OT128 User Manual](https://www.hesaitech.com/wp-content/uploads/2025/03/OT128_User_Manual_O01-en-250310.pdf).

Basically, the pattern is repeating every 2 blocks x 4 channels, so the specialized transform ensures that the points that change within one tile are mapped to the same index for dithering.

#### Filter disabled (unaffected by this PR)
The dithering pattern causes some oscillating points but the image is overall stable.

https://github.com/user-attachments/assets/9375efdc-9084-4331-a940-005df1778b19 
#### Filter enabled, before fix 
The interference between sensor dithering and downsample filter causes strong flickering.

https://github.com/user-attachments/assets/b005ad64-b256-4d39-bdb2-01c81930b2e2 
#### Filter enabled, after fix 
The specialized dithering pattern plays well with the sensor dithering.
Osillations are more noticeable than in the disabled case due to there being larger gaps between points.
The actual oscillation amplitude is the same.

https://github.com/user-attachments/assets/107d2bc2-ffbb-4432-9367-f1caef1e9402 

## Review Procedure

* Code review
* Compare implementation with datasheet
* Test with downsample mask (different downsample factors or gradients)

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
+++ b/nebula_decoders/include/nebula_decoders/nebula_decoders_common/point_filters/downsample_mask.hpp
@@ -31,6 +31,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <filesystem>
+#include <functional>
 #include <memory>
 #include <sstream>
 #include <stdexcept>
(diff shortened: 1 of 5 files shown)


drwnz (approved):
Code looks good, and I could confirm the issue with this rosbag: [TIER IV INTERNAL LINK](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/282802b7-dc1e-4327-9516-43d9699b55ad), and also that the flicker was mitigated using this PR.

PR #291, opened 2025-03-21
Hot fix for VLP16 timestamp

ike-kazu:
## PR Type

Only vlp16 has problems for timestamp. In this PR, hot fix is created for this.

- Bug Fix

## Related Links
[VLP16 timestamp problem report
](https://tier4.atlassian.net/wiki/spaces/CT/pages/3607168528/VLP16+timestamp+problem+report?force_transition=7657d3d6-104a-4f92-8d0c-67bddab81400)

## Description

Only VLP16 has problems for timestamps in poincloud from nebula, **Last Filtered Pointcloud timestamp problem**.
When vlp16 pointcloud is filtered through nebula, last pointcloud timestamp is weird sometime. We research about this problem but we can't discover the cause of this clearly. 
But we Discover that VLP16 pointcloud timestamps are weird when it is filtered by hardware.
It's fine when VLP16 pointcloud are filtered by only nebula.

From this discovery, we decide that we create hardware filter offset as hot fix for this problem.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp
+++ b/nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp
@@ -167,21 +167,44 @@ VelodyneStatus VelodyneHwInterface::check_and_set_config(
               << std::endl;
   }
 
+  int setting_cloud_min_angle = sensor_configuration->cloud_min_angle;
+  int setting_cloud_max_angle = sensor_configuration->cloud_max_angle;
+
+  // FIXME: VLP16 has problems for timestamp. Whatch github issue #
+  if (sensor_configuration->sensor_model == SensorModel::VELODYNE_VLP16) {
+    int angle_diff = (setting_cloud_max_angle - setting_cloud_min_angle + 360) % 360;
+
+    if (angle_diff >= 360 || angle_diff == 0) {
+      setting_cloud_min_angle = 0;
+      setting_cloud_max_angle = 359;
+    } else {
+      if (setting_cloud_min_angle < 5) {
+        setting_cloud_min_angle = (setting_cloud_min_angle + 355) % 360;
+      } else {
+        setting_cloud_min_angle -= 5;
+      }
+
+      if (setting_cloud_max_angle > 354) {
+        setting_cloud_max_angle = 359;
+      } else {
+        setting_cloud_max_angle += 5;
+        if (setting_cloud_max_angle >= 360) {
+          setting_cloud_max_angle = 359;
+        }
+      }
+    }
+
+  } else {
+    if (setting_cloud_min_angle == 360) {
+      setting_cloud_min_angle = 359;
+    }
+    if (setting_cloud_max_angle == 360) {
+      setting_cloud_max_angle = 359;
+    }
+  }
+
   target_key = "config.fov.start";
   auto current_cloud_min_angle = tree.get<std::uint16_t>(target_key);
-  int setting_cloud_min_angle = sensor_configuration->cloud_min_angle;
-  // // FIXME
-  // if (sensor_configuration->sensor_model == SensorModel::VELODYNE_VLP16) {
-  //   if (setting_cloud_min_angle < 5){
-  //     setting_cloud_min_angle = setting_cloud_min_angle + 360 - 5;
-  //   }else{
-  //     setting_cloud_min_angle -= 5;
-  //   }
-  // }
-  // Velodyne only allows a maximum of 359 in the setting
-  if (setting_cloud_min_angle == 360) {
-    setting_cloud_min_angle = 359;
-  }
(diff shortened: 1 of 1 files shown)


nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp
```diff
@@ -167,13 +167,44 @@ VelodyneStatus VelodyneHwInterface::check_and_set_config(
...
+
+    if (angle_diff >= 360 || angle_diff == 0) {
+      setting_cloud_min_angle = 0;
+      setting_cloud_max_angle = 359;
+    } else {
+      if (setting_cloud_min_angle < 5) {
+        setting_cloud_min_angle = (setting_cloud_min_angle + 355) % 360;
+      } else {
+        setting_cloud_min_angle -= 5;
+      }
```
mojomex, line 185: This if/else can be simplified into one generic formula (`-5 + 360` might be more understandable than `+355`):
```suggestion
      setting_cloud_min_angle = (setting_cloud_min_angle - 5 + 360) % 360;
```

nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp
```diff
@@ -167,13 +167,44 @@ VelodyneStatus VelodyneHwInterface::check_and_set_config(
...
+
+      if (setting_cloud_max_angle > 354) {
+        setting_cloud_max_angle = 359;
+      } else {
+        setting_cloud_max_angle += 5;
+        if (setting_cloud_max_angle >= 360) {
+          setting_cloud_max_angle = 359;
+        }
+      }
```
mojomex, line 194: Can be simplified to:
```suggestion
      setting_cloud_max_angle = (setting_cloud_max_angle + 5) % 360;
```

nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp
```diff
@@ -167,13 +167,44 @@ VelodyneStatus VelodyneHwInterface::check_and_set_config(
...
 
-  target_key = "config.fov.start";
-  auto current_cloud_min_angle = tree.get<std::uint16_t>(target_key);
   int setting_cloud_min_angle = sensor_configuration->cloud_min_angle;
-  // Velodyne only allows a maximum of 359 in the setting
-  if (setting_cloud_min_angle == 360) {
-    setting_cloud_min_angle = 359;
+  int setting_cloud_max_angle = sensor_configuration->cloud_max_angle;
+
+  // FIXME: VLP16 has problems for timestamp. Whatch github issue #
```
mojomex, line 173: Please explain in more detail (combination of hardware and software scan cutting causes timestamping issues at the FoV borders, so the hardware FoV has to be increased for scan cutting to work.

nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp
```diff
@@ -167,13 +167,44 @@ VelodyneStatus VelodyneHwInterface::check_and_set_config(
...
+  if (sensor_configuration->sensor_model == SensorModel::VELODYNE_VLP16) {
+    int angle_diff = (setting_cloud_max_angle - setting_cloud_min_angle + 360) % 360;
```
mojomex, line 175: Please define the angle we oversize by as a `const int32_t fov_tolerance_deg = 5;` and replace all `5`s below by that constant.

nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp
```diff
@@ -167,13 +167,44 @@ VelodyneStatus VelodyneHwInterface::check_and_set_config(
...
+
+    if (angle_diff >= 360 || angle_diff == 0) {
```
mojomex, line 177: Angle diff cannot be `>= 360` (it was `% 360`ed), so the below condition is sufficient:
```suggestion
    if (angle_diff == 0) {
```

mojomex (changes requested):
Thank you for the PR!
I have left a few comments about simplifying the code, please have a look.

Also, could you add a self-evaluation section where you use a real VLP16 with different settings?

Ones I'm interested in are:
* 0, 360 -- no gaps should be observed
* 90, 90 -- should be the same output as 0, 360
* 0, 180 -- no weird timestamp0s at edges
* 180, 0 -- same but the other half of the pointcloud is shown
* 355, 5 -- only 10deg of the pointcloud are shown
* 5, 355 -- a 10deg gap is shown
* 3, 357 -- a 6deg gap is shown

PR #295, opened 2025-03-25
chore(hires_mode): amend #294 - add forgotten-to-commit config option and config print

mojomex:
See #294. Forgot to push the config default and the config print for hires mode :bow:


mojomex:
(results shown in #294 already, merging myself)

PR #277, opened 2025-02-20
perf(robosense): remove mtqueue

mojomex:
## PR Type

- Improvement
- Bug fix

## Related Links

- #226 -- Hesai for reference
- #240 -- smoke tests for reference

## Description

Does what the title says.

Reasons for this change:
* The multithreaded queue is not needed (receive buffers in the network stack take care of arriving data while decoding).
* The decoder thread had no mechanism for gracefully stopping, thus causing an exception on exit
* The above excption caused our smoke tests to fail

## Review Procedure

* Compare to the linked Hesai reference PR #226
* Test with `launch_hw:=false` and a rosbag
* Test with `launch_hw:=true` and a real sensor or pcap

## Pre-Review Checklist for the PR Author

**PR Author should check the checkboxes below when creating the PR.**

- [ ] Assign PR to reviewer

## Checklist for the PR Reviewer

**Reviewers should check the checkboxes below before approval.**

- [ ] Commits are properly organized and messages are according to the guideline
- [ ] (Optional) Unit tests have been written for new behavior
- [ ] PR title describes the changes

## Post-Review Checklist for the PR Author

**PR Author should check the checkboxes below before merging.**

- [ ] All open points are addressed and tracked via issues or tickets

## CI Checks

- **Build and test for PR**: Required to pass before the merge.

PR #296, opened 2025-03-25
chore: remove unused mt_queue

mojomex:
The `mt_queue.hpp` header became unused after merging #274, #275, #276, #277. Delete it to keep dead code to a minimum.


mojomex:
Since this is a trivial change, I will merge it myseld

PR #304, opened 2025-04-10
chore(docs): ensure word breaks in long paths/URLs so that tables don't overflow

mojomex:
## PR Type

- Improvement

## Description

Markdown tables generated from parameter schemas were overflowing due to parameter defaults like `$(find-pkg-share nebula_decoders)/calibration/hesai/$(var sensor_model).csv` which don't have any word breaks.

While the parameters cannot really be changed/shortened, this PR introduces an improvement to the `json_to_markdown` macro which adds [HTML `<wbr>` (line break opportunity) tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) after all `/`es in default parameters. If a parameter value is short enough to fit, this does not have any effect. For longer paths like the one above, line breaks can be added by the browser at those `<wbr>` marks, causing the table to render correctly.

| Before | After |
|-|-|
| Screenshot from 2025-04-10 12-00-26 | Screenshot from 2025-04-10 12-00-52 |

## Review Procedure

* confirm the tables at https://tier4.github.io/nebula/parameters/vendors/hesai/ot128/ are broken for you
* build and view docs locally with `mkdocs serve` and see whether the issue is fixed

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/mkdocs_macros.py
+++ b/mkdocs_macros.py
@@ -18,6 +18,14 @@ def format_param_type(param_type):
         return param_type
 
 
+def ensure_word_breaks(text):
+    """For URLs/paths, insert HTML word break marks. Does not affect other or non-string inputs."""
+    if not isinstance(text, str):
+        return text
+
+    return text.replace("/", "/<wbr>")
+
+
 def format_param_range(param):
     list_of_range = []
     if "enum" in param.keys():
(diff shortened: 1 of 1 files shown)


knzo25:
Can you explain briefly the checks needed?

mojomex:
@knzo25 I added them to the PR description now :bow:

knzo25 (approved):
LGTM

PR #307, opened 2025-04-22
refactor(hesai): centralized diagnostic_updater

mojomex:
## PR Type

- Improvement

## Description

This PR moves the diagnostic_updater from hw_monitor to ros_wrapper for Hesai.
Since the updater will be used by various diagnostics not just in HW monitor, this change allows for clean implementation of those coming-soon diagnostics.

Other effects:

* The redundant force_update timer has been removed, as diagnostic_updater already has its own timer.
* A get_inventory has been sent to the sensor from 2 places before, this has been reduced to 1.

Confirmed working with a [dummy TCP connection](https://github.com/tier4/pcap_replay) and a real sensor. Publish rate is the same 1 Hz as before.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/src/hesai/hesai_ros_wrapper.cpp
+++ b/nebula_ros/src/hesai/hesai_ros_wrapper.cpp
@@ -96,9 +96,7 @@ HesaiRosWrapper::HesaiRosWrapper(const rclcpp::NodeOptions & options)
       "Hardware connection disabled, listening for packets on " << packets_sub_->get_topic_name());
   }
 
-  auto hardware_id = hw_interface_wrapper_ && hw_interface_wrapper_->inventory()
-                       ? hw_interface_wrapper_->inventory()->to_hardware_id()
-                       : "none";
+  auto hardware_id = sensor_cfg_ptr_->frame_id;
 
   diagnostic_updater_.setHardwareID(hardware_id);
 
--- a/nebula_ros/src/hesai/hesai_ros_wrapper.cpp
+++ b/nebula_ros/src/hesai/hesai_ros_wrapper.cpp
@@ -28,7 +28,7 @@ HesaiRosWrapper::HesaiRosWrapper(const rclcpp::NodeOptions & options)
 : rclcpp::Node("hesai_ros_wrapper", rclcpp::NodeOptions(options).use_intra_process_comms(true)),
   wrapper_status_(Status::NOT_INITIALIZED),
   sensor_cfg_ptr_(nullptr),
-  diagnostic_updater_(this)
+  diagnostic_updater_((declare_parameter<bool>("diagnostic_updater.use_fqn", true), this))
 {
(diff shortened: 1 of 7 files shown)


nebula_ros/src/hesai/hesai_ros_wrapper.cpp
```diff
@@ -97,6 +96,12 @@ HesaiRosWrapper::HesaiRosWrapper(const rclcpp::NodeOptions & options)
       "Hardware connection disabled, listening for packets on " << packets_sub_->get_topic_name());
   }
 
+  auto hardware_id = hw_interface_wrapper_ && hw_interface_wrapper_->inventory()
```
drwnz, line 99: This is where it is potentially going to be more useful to use the device name (from the frame_id?)

nebula_ros/src/hesai/hw_monitor_wrapper.cpp
```diff
@@ -53,28 +51,19 @@ HesaiHwMonitorWrapper::HesaiHwMonitorWrapper(
...
-void HesaiHwMonitorWrapper::initialize_hesai_diagnostics(bool monitor_enabled)
+void HesaiHwMonitorWrapper::initialize_hesai_diagnostics(
+  diagnostic_updater::Updater & diagnostic_updater, bool monitor_enabled)
 {
   using std::chrono_literals::operator""s;
   std::ostringstream os;
-  diagnostics_updater_.add("hesai_status", this, &HesaiHwMonitorWrapper::hesai_check_status);
-  diagnostics_updater_.add("hesai_ptp", this, &HesaiHwMonitorWrapper::hesai_check_ptp);
-  diagnostics_updater_.add(
+  diagnostic_updater.add("hesai_status", this, &HesaiHwMonitorWrapper::hesai_check_status);
```
drwnz, line 62: Since our design now has human-readable task names, it would be good to align for these items.

drwnz (commented):
Looks good, just some comments about syntax and hardware ID.

nebula_ros/src/hesai/hw_monitor_wrapper.cpp
mojomex: We should probably also combine them into a single `DiagnosticStatus`, rather than having four separate ones.
The key-value pairs should also receive updated names, so I'd suggest we do this in a separate PR?

drwnz: Agreed :+1:

nebula_ros/src/hesai/hesai_ros_wrapper.cpp
mojomex: Changed to frame ID (see below). This has the added benefit of also working in offline replay :partying_face: 
image

drwnz (approved):
LGTM!

Issue #310, opened 2025-04-24
Velodyne: No pointclouds in udp_only mode unless host_ip is set to broadcast_ip


mbharatheesha:
When using `udp_only=true` for a VLP16 usecase, I notice that pointclouds are only available on `/velodyne_points` only if `host_ip=255.255.255.255`. If it is set to the actual `host_ip`, no pointclouds are published.

However, with a Hesai LiDAR with `udp_only=true` and a specific `host_ip`, pointclouds are published to `pandar_points`. In fact, if `host_ip` is set to a broadcast ip address, the driver errors out as expected.

mojomex:
Thanks for the feedback!

That's interesting. The real host IP and `0.0.0.0` should definitely work. `255.255.255.255` should be disallowed, but such checks are not implemented in `transport_drivers`.

This is probably the right time to transition to the new UDP socket implementation that is independent of `transport_drivers` (see #235).

I can do that pretty quickly, but would you be willing to test my PR with a real device? I don't have one to check for the next three weeks.

mbharatheesha:
@mojomex I can definitely test this if the fix is implemented.. Thank you for the quick response.

drwnz:
@mbharatheesha did this solve your issue?

mbharatheesha:
@drwnz Yes, this issue is fixed. Sorry I didn't close it sooner.

PR #312, opened 2025-04-25
perf(velodyne): use new UDP socket implementation

mojomex:
## PR Type

- Improvement

## Related Links

- #231 -- using the UDP socket from this PR
- #235 -- equivalent PR for Hesai

## Description

Use the new UDP socket introduced in #231 as a drop-in, ASIO-free, transport-driver-free replacement for the old UdpDriver in HesaiHwInterface.

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/config/lidar/velodyne/VLP16.param.yaml
+++ b/nebula_ros/config/lidar/velodyne/VLP16.param.yaml
@@ -1,6 +1,6 @@
 /**:
   ros__parameters:
-    host_ip: 255.255.255.255
+    host_ip: 192.168.1.10
     sensor_ip: 192.168.1.201
     data_port: 2368
     gnss_port: 2369
--- a/nebula_hw_interfaces/include/nebula_hw_interfaces/nebula_hw_interfaces_velodyne/velodyne_hw_interface.hpp
+++ b/nebula_hw_interfaces/include/nebula_hw_interfaces/nebula_hw_interfaces_velodyne/velodyne_hw_interface.hpp
@@ -53,10 +53,11 @@ namespace nebula::drivers
 class VelodyneHwInterface
 {
 private:
-  std::optional<connections::UdpSocket> udp_socket_;
   std::shared_ptr<const VelodyneSensorConfiguration> sensor_configuration_;
-  std::function<void(const std::vector<uint8_t> &)>
-    cloud_packet_callback_; /**This function pointer is called when the scan is complete*/
+  /**This function pointer is called when the scan is complete*/
+  std::function<void(const std::vector<uint8_t> &)> cloud_packet_callback_;
+  // Calls the above `cloud_packet_callback_` and thus has to be destroyed before it.
+  std::optional<connections::UdpSocket> udp_socket_;
 
   std::shared_ptr<boost::asio::io_context> boost_ctx_;
   std::unique_ptr<::drivers::tcp_driver::HttpClientDriver> http_client_driver_;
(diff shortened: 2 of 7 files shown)


mojomex:
@mbharatheesha This is the new UDP implementation mentioned in #310.
It seems I can't add you as a reviewer but If you find the time, please see if the issue persists despite this PR.

I did test with a PCAP replayed on localhost, but due to it being an emulated sensor connection, behavior might be different with a real one.

mbharatheesha:
@mojomex Thanks for the changes and support. I just tested these with a VLP16. A few comments:

1. These lines now need to all change to a valid `host_ip` ([VLP16](https://github.com/mojomex/nebula/blob/df7c1ce40647c41d904368537ddea97316149895/nebula_ros/config/lidar/velodyne/VLP16.param.yaml#L3),[VLP32](https://github.com/mojomex/nebula/blob/df7c1ce40647c41d904368537ddea97316149895/nebula_ros/config/lidar/velodyne/VLP32.param.yaml#L3), [VLP128](https://github.com/mojomex/nebula/blob/df7c1ce40647c41d904368537ddea97316149895/nebula_ros/config/lidar/velodyne/VLP128.param.yaml#L3))
2. When the source is compiled _without_ `--cmake-args -DCMAKE_BUILD_TYPE=Release`, I get a segfault that seemingly originates from [here](https://github.com/mojomex/nebula/blob/df7c1ce40647c41d904368537ddea97316149895/nebula_hw_interfaces/src/nebula_velodyne_hw_interfaces/velodyne_hw_interface.cpp#L68)
  - I can confirm this also happens on `main` without the changes in this PR.
  - I can also confirm that this happens with a Hesai LiDAR too.
  - This happens irrespective of whether `udp_only` is set to `true` or `false`

And the testing results when compiled with `--cmake-args -DCMAKE_BUILD_TYPE=Release`:
- When `host_ip: 255.255.255.255`, I get a runtime error as expected
- When I use a valid `host_ip` such as `192.168.1.10`, I get pointclouds as expected in both configurations, i.e., `udp_only: true` and `udp_only: false`

So, the changes are working as intended. Perhaps, the segfault in the non-release build is an unrelated issue to this PR.

mbharatheesha:
> As for the segfault, I can't reproduce it without a sensor, but have historically almost only been compiling with RelWithDebInfo. If you have any stack trace it would be greatly appreciated.

Of course, sorry that I didn't paste it already earlier!
image

mojomex:
@mbharatheesha
Thank you!

I was able to reproduce the issue when compiling in `Debug` mode. While the code did not segfault on m machine, the `recvmsg` call returned an `errno 14: bad address`. This is an error returned by the kernel to signal an invalid pointer passed to a syscall. After some debugging, I found that `make_msg_headers` in the `UdpDriver` was faulty and returned an object containing dangling pointers. Fixed in #313.

That PR is stacked on top of this one, so if you find the time, could you confirm that the segfault does not occurr with that fix anymore?

Thank you for your continued support!

mbharatheesha:
@mojomex Thank you for the fixes.

I confirm that the changes in #313 fixes the segfault issue. This was tested with a Velodyne and a Hesai LiDAR and everything works as expected when `udp_only: true` and `udp_only: false`.

mojomex:
PR #313 has been merged to main, fixing the segfault issue. This PR has been rebased onto the newest main, so everything is expected to work as tested [here](https://github.com/tier4/nebula/pull/312#issuecomment-2838335807).

@drwnz could you take a quick look?

nebula_ros/config/lidar/velodyne/VLP16.param.yaml
```diff
@@ -1,6 +1,6 @@
 /**:
   ros__parameters:
-    host_ip: 255.255.255.255
+    host_ip: 192.168.1.10
```
drwnz, line 3: Very minor point, but all Velodyne datasheets seem to use 192.168.1.222 in the example setup. Although some additionally use 192.168.1.77 to show how to set up multiple sensors, so using the same as the user manual isn't really going to make things easier to understand anyway...

mbharatheesha:
@mojomex Will this be merged soon? If so, I will base my Lifecycle node changes on top of this and I can create a PR with that. Of course my assumption is that Lifecycle implementations will be the way forward. If not, I will just take these changes over on my development branch.

mojomex:
@mbharatheesha Sorry for the delay, it's merged now 🙌

PR #273, opened 2025-02-20
feat(ars548): use new UDP socket

mojomex:
## PR Type

- Improvement

## Related Links

- #231 -- introduces the new UDP socket (receive-only)
- #272 -- provides the send functionality needed for this PR

## Description

Introduce the new UDP socket to the ARS548 hardware interface. This 
* allows for accurate ingress timestamping
* will allow us to phase out transport_drivers in the future (#267)

## Review Procedure

Test with a real sensor if all data is sent/received as expected.

## Pre-Review Checklist for the PR Author

**PR Author should check the checkboxes below when creating the PR.**

- [ ] Assign PR to reviewer

## Checklist for the PR Reviewer

**Reviewers should check the checkboxes below before approval.**

- [ ] Commits are properly organized and messages are according to the guideline
- [ ] (Optional) Unit tests have been written for new behavior
- [ ] PR title describes the changes

## Post-Review Checklist for the PR Author

**PR Author should check the checkboxes below before merging.**

- [ ] All open points are addressed and tracked via issues or tickets

## CI Checks

- **Build and test for PR**: Required to pass before the merge.

PR #313, opened 2025-04-28
fix(udp): fix dangling pointers in `MsgBuffers`

mojomex:
## PR Type

- Bug Fix

## Related Links

- #312 -- detected here
- #273 -- also affects this
- #235 -- also affects this

## Description

The way the `make_msg_header` function worked, it assigned pointers to local variables in a `msghdr` instance that was then returned. Those pointers thus became dangling, causing the kernel to return an `errno 14: bad address` and crashing Nebula.

This fix deals with correct assignment in the `MagBuffers` constructor, the pointers pointing to the initialized instance's member variables. The pointers thus remain valid over the `MsgBuffers` object's lifetime.

## Review Procedure

- Compile Nebula with `-DCMAKE_BUILD_TYPE=Debug`
- Connect to a sensor (or replay a PCAP). `udp_only` mode is fine but a full connection works too
- Confirm that the issue happens without this PR, but is solved without this PR.

- [ ] Assign PR to reviewer

## CI Checks


mojomex:
See https://github.com/tier4/nebula/pull/312#issuecomment-2838335807 for hardware test results.
Since this is an obvious bug, has been confirmed fixed in the above comment and it is currently blocking some other PRs, I am merging this without waiting for a review.

PR #320, opened 2025-05-13
fix: update message_filters includes for compatibility with old environments

technolojin:
## PR Type
- Bug Fix

## Related Links
https://github.com/autowarefoundation/autoware_core/pull/467

## Description
In some old environment, the build is failed.
Similar to the related link, resolve the issue by macro.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/include/nebula_ros/continental/continental_srr520_hw_interface_wrapper.hpp
+++ b/nebula_ros/include/nebula_ros/continental/continental_srr520_hw_interface_wrapper.hpp
@@ -16,9 +16,16 @@
 
 #include "nebula_ros/common/parameter_descriptors.hpp"
 
+#if __has_include(<message_filters/subscriber.hpp>)
 #include <message_filters/subscriber.hpp>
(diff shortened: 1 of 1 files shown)


mojomex:
Exactly for cases like this, I don't like the way the ROS ecosystem handles dependency versioning (it basically doesn't enforce anything). Instead, the user is made responsible for keeping their environment up to date.

Since this is the norm in the ecosystem, I think we should stick to it (especially given that any new builds of Autoware require rosdep to be up-to-date anyway and that this is a very temporary issue).

For that reason, I'd like to not add extra boilerplate to keep compatible with outdated revisions of Humble.
@drwnz What's your take on this?

technolojin:
@mojomex 
I agree. That's reason why I kept it as a draft.

I updated my environment and could run the system well. This is what we should go.

Related PR 
Fix header for Universe https://github.com/autowarefoundation/autoware_universe/pull/10586#issuecomment-2866538495

PR #322, opened 2025-05-16
Feat/cloud watchdog diag

mojomex:
## PR Type

- New Feature

## Description

This PR introduces a second, 10 Hz diagnostic updater into the Hesai ROS wrapper. The pointcloud publish rate is monitored with a newly implemented `FrequencyDiagnosticTask` and published using the new updater.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/include/nebula_ros/hesai/decoder_wrapper.hpp
+++ b/nebula_ros/include/nebula_ros/hesai/decoder_wrapper.hpp
@@ -16,8 +16,11 @@
 
 #include "nebula_decoders/nebula_decoders_hesai/hesai_driver.hpp"
 #include "nebula_hw_interfaces/nebula_hw_interfaces_hesai/hesai_hw_interface.hpp"
+#include "nebula_ros/common/diagnostics/frequency_diagnostic_task.hpp"
 #include "nebula_ros/common/watchdog_timer.hpp"
 
+#include <diagnostic_updater/publisher.hpp>
+#include <diagnostic_updater/update_functions.hpp>
 #include <nebula_common/hesai/hesai_common.hpp>
 #include <nebula_common/nebula_common.hpp>
 #include <rclcpp/rclcpp.hpp>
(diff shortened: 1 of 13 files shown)


nebula_ros/include/nebula_ros/hesai/decoder_wrapper.hpp
```diff
@@ -64,6 +67,29 @@ class HesaiDecoderWrapper
       std::chrono::duration<double>(seconds));
   }
 
+  static custom_diagnostic_tasks::RateBoundStatus make_rate_bound_status(
```
drwnz, line 70: For these, is there a need to convert to relative for the parameters?
It may be more intuitive to read diagnostics that are 1:1 with the parameters, as per the camera rate diagnostics PR.

drwnz (commented):
This looks good, but open to discussion on the parametrization. As mentioned, I think I prefer the parameters being min and max acceptable rates rather than percentages, as in the field it's a bit more natural to check the diagnostics results and tune the parameters. Or is there a particular case for using the relative parametrization?

nebula_ros/include/nebula_ros/hesai/decoder_wrapper.hpp
mojomex: The thought was that the absolute value is given by whatever RPM is set, and that the user has to calculate in their head (e.g. AT128: 200 RPM -> 10 Hz, QT128: 600 RPM -> 10 Hz, 1200 RPM -> 20 Hz, etc.).
Additionally, if it's absolute, the user now has to change 5 parameters instead of one when changing RPM. What do you think?

drwnz: It's unlikely that running at 20Hz as opposed to 10Hz would still follow the same linear pattern for bounds anyway (i.e, 10% margin would be 1Hz for 600rpm but 2Hz at 1200rpm - in reality I don't think higher frame rate = setting higher absolute margin).
Thinking about how this diagnostic would actually be used - I imagine the sensor would be setup, a rough value for max and min put in, and then adjusted depending one the actual operation in the vehicle - where the people doing the tuning would be unlikely to calculate based on RPM and more likely to look at the diagnostic result and then change the margin parameters to suit, based on the actual frame rate achieved.
@manato do you have an opinion on this?

mojomex:
Updated the video above to include the latest changes.

PR #325, opened 2025-05-21
feat(hesai): packet loss diagnostics

mojomex:
## PR Type

- Improvement

## Related Links

- [TIER IV INTERNAL](https://tier4.atlassian.net/browse/RT1-5636) -- Jira task

## Description

Adds diagnostics for packet loss based on the `UDP Sequence` field that some Hesai sensors have.
If the field did not increase by exactly `1` (or roll over from `0xFF FF FF FF` to `0`), the difference of the actual and expected sequence number is treated as the number of lost packets:

$$ n_{lost;i} = seq_{i} - 1 - seq_{i-1} $$

e.g.

$$ n_{lost;i} = 2005 - 1 - 2000 = 4 $$

These are accumulated for each `100 ms` diagnostic cycle (which might be running at a different phase than pointcloud output!) and the accumulated sum is used for diagnostic output:

$$ lost = \sum_i lost_i $$

```python
if lost == 0:
  status = OK
elif lost < param("diagnostics.packet_loss.error_threshold"):
  status = WARN
else:
  status = ERROR
```

### Here it is in action:

[Screencast from 05-27-25 21_39_39.webm](https://github.com/user-attachments/assets/3e1729fb-5ffe-4caa-a94a-4e0f5d67f560)

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_ros/include/nebula_ros/hesai/diagnostics/packet_loss_diagnostic.hpp
+++ b/nebula_ros/include/nebula_ros/hesai/diagnostics/packet_loss_diagnostic.hpp
@@ -43,6 +43,7 @@ public:
     last_run_time_ = now;
 
     status.add("Lost packets", n_lost_packets_);
+    status.add("Error threshold", error_threshold_);
     status.add("Time since last update [s]", std::to_string(dt.seconds()));
 
     if (n_lost_packets_ == 0) {


drwnz (changes requested):
Looks good!
Because we don't really know what levels of packet loss are really serious at this stage, it would be good to expose the WARN and ERROR parameters to the diagnostic status keys.

mojomex:
Here's the requested threshold info, with `error_threshold = 100` and simulated stochastic packet loss ~100 packets / scan:

[Screencast from 2025年05月30日 17時59分58秒.webm](https://github.com/user-attachments/assets/529bc45e-3b71-4a31-9340-8b6fb4fb905e)

drwnz (approved):
LGTM!

PR #336, opened 2025-06-17
fix(ars548): re-introduce correct UDP MTU and sender filter

mojomex:
## PR Type

- Bug Fix

## Related Links

- #273 -- introduced the bug here :bow: 

## Description

This PR re-introduces the sender filter for ARS548. As the sensor is using IP multicast, if there are multiple ARS548s in the same IP subnet, this leads to Nebula receiving data from each of them without this filter, which is unwanted behavior. 

The PR also re-introduces the MTU of `2^16 B` without which the large detection list / object list datagrams cannot be received by Nebula.

- [ ] Assign PR to reviewer

## CI Checks


amadeuszsz (approved):
> # 🟢 Self-Evaluation
> The MTU size was also part of the problem. The sender filter was confirmed working on the bench together with @amadeuszsz, however, we don't have video footage 🙇
> 
>  ars548-udp-fix.mp4


@mojomex 
Thanks again! Looks it works, LGTM!

Issue #339, opened 2025-07-02
Hesai LiDAR sometimes continues publishing empty pointclouds when cut_angle=cloud_max_angle!=360


KYabuuchi:
## Conditions
Launch XT32 with nebula v0.2.5.1 using the following parameters:
``` yaml
  sync_angle: 310
  cut_angle: 310.0
  cloud_min_angle: 40
  cloud_max_angle: 310
```

## Problem
I encountered an issue where XT32 occasionally publishes empty pointclouds (`pointcloud_raw_ex`). Once this happens, restarting nebula does not resolve the problem.  
This issue does not occur with only a XT32 sensor, but has been observed with some XT32 sensors, so it does not seem to be a sensor-specific problem.  
Also, basically it publishes correct (not empty) pointclouds.

<img src=https://github.com/user-attachments/assets/dad3fdfb-bc0f-4c22-bd01-a13467293bc8 width=400>

## My Analysis
From debug prints, I found that when an empty pointcloud is published, the accumulated pointcloud data is being reset [here](https://github.com/tier4/nebula/blob/v0.2.5.1/nebula_decoders/include/nebula_decoders/nebula_decoders_hesai/decoders/hesai_decoder.hpp#L315).
It seems there may be an issue where packets around `cloud_max_angle` continuously fail to be received. 🤔 

## Links
* [TIER IV INTERNAL DISCUSSION](https://star4.slack.com/archives/C076E8EBJTG/p1750742865586399)
* [TIER IV INTERNAL TICKET](https://tier4.atlassian.net/browse/RT0-37996?atlOrigin=eyJpIjoiNGEzYzljNjg4NDQyNGNiMGI0MTY5NTQ4MDY5MGU0OTUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

mojomex:
Fixed by #353

PR #353, opened 2025-08-25
fix(hesai_hw_interfaces): slightly widen FoV

KYabuuchi:
## PR Type

- Bug Fix

## Related Links

* https://github.com/tier4/nebula/issues/339

## Description

Based on my experience, the Hesai XT32 occasionally does not publish packets near the boundary of the FOV.
When that happens, Nebula cannot publish the point cloud at the correct timing. :cry: 

Currently, Nebula uses a calibration file downloaded to set the minimum FOV for the LiDAR, but this seems to be not sufficient. **Therefore, I added a fix that adjusts the FOV by ~~0.1~~ `1.0 degrees` .**

> [!NOTE]
> At least for all the “problematic” LiDARs I have encountered, this fix has been sufficient.

### My Experiments

Here is a video of one of the “problematic” LiDARs I encountered.

https://github.com/user-attachments/assets/ef7b711b-031f-4551-858b-668d7660c1f3

When running nebula (v0.2.6.1), only a single line of point clouds is published.
However, if I configure the LiDAR via the web interface to use an FoV that is wider by just 0.1 degrees, the expected point clouds are displayed.

When Nebula is restarted, the LiDAR is reset to the minimum FoV, and once again only a strange point cloud is published.
With my PR applied, this issue is resolved.

## Review Procedure

~~I have confirmed that the FOV of the XT32 expands by 0.1 degrees.~~
I have confirmed that the FOV of the XT32 expands by 1.0 degrees.

## Remarks

Nothing

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_common/include/nebula_common/hesai/hesai_common.hpp
+++ b/nebula_common/include/nebula_common/hesai/hesai_common.hpp
@@ -202,6 +202,12 @@ struct HesaiCalibrationConfiguration : public HesaiCalibrationConfigurationBase
       max = std::max(max, item.second);
     }
 
+    // NOTE: Slightly widen the FOV padding because some LiDARs do not transmit blocks near the end
+    // of the FOV. If these blocks are missing, the point cloud may not be published at the desired
+    // timing.
+    min -= 0.1f;
+    max += 0.1f;
+
     return {-max, -min};
   }
 };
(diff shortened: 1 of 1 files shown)


mojomex (approved):
Thank you!
As discussed, this is a slightl hacky solution, but more robust than rounding up/down the FoV calculated from the calibration values. 

Especially in the face of packet loss, I think having this 1deg buffer on each side makes behavior more stable, so I'll approve 👍

PR #357, opened 2025-09-22
fix(continental_ars548): objects list specs compatibility

amadeuszsz:
## PR Type

- Improvement

## Related Links

[TIER IV INTERNAL LINK](https://star4.slack.com/archives/CEV8XMJBV/p1758275439614179?thread_ts=1756443528.624719&cid=CEV8XMJBV)

## Description

Allow multiple object list specifications. ~~Behavior is not bound to the firmware version~~. The data format is inferred based on the received packet's UDP payload and PDU length.

### With vehicle
1. Build package.
```bash
mkdir -p ~/nebula_fix_ws/src
git clone https://github.com/amadeuszsz/nebula --branch fix/objects-list-compatibility ~/nebula_fix_ws/src/nebula
git clone https://github.com/autowarefoundation/autoware_msgs ~/nebula_fix_ws/src/autoware_msgs
git clone https://github.com/tier4/sync_tooling_msgs.git ~/nebula_fix_ws/src/sync_tooling_msgs.git
cd ~/nebula_fix_ws
source /opt/ros/humble/setup.bash
source ~/pilot.auto.xx1/install/setup.bash
colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=On
```

2. Add `source ~/pilot.auto.xx1/install/setup.bash` to `~/autoware.env` one line after `source /opt/ros/humble/setup.bash`.

3. Make sure `ARS548.param.yaml` from current `aip_launcher` consists of recent parameters ([radar_info_rate_subsample](https://github.com/tier4/aip_launcher/blob/tier4/universe/aip_xx1_gen2_launch/config/ARS548.param.yaml#L14) and [diagnostics](https://github.com/tier4/aip_launcher/blob/tier4/universe/aip_xx1_gen2_launch/config/ARS548.param.yaml#L19-L33)). If they are missing, add them in order to test new nebula.

4. Run Autoware service.

## Remarks

- Corner RADAR check has been removed. This is required for corner radars with dedicated corner-RADAR firmware. We need to double-check behavior for corner radars with non-corner-RADAR firmware.
- After validation with hardware, backport to `0.2.4.X`.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/nebula_common/include/nebula_common/continental/continental_ars548.hpp
+++ b/nebula_common/include/nebula_common/continental/continental_ars548.hpp
@@ -22,11 +22,10 @@
 #include <pcl/point_cloud.h>
 #include <pcl/point_types.h>
 
-#include <array>
 #include <iostream>
 #include <optional>
 #include <string>
-#include <variant>
+#include <vector>
 
 namespace nebula
 {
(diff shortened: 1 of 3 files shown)


drwnz:
Thank you at @amadeuszsz !
Although the corner radar check has been removed, I think we really need to check the firmware version for this. The reason being, that we will still have some pre-updated devices being used as corner radars which still need the object list parsing disabled, as before. While I suppose this could be tested using the object packet size and radar's yaw setting, if there is a facility to check the firmware version this may be simpler.

@shmpwk can we ask you to test this branch? I believe it should work fine for the taxi, and the above comment is only relevant for  devices without the firmware change in corner radars.
I think before we merge we will also have to test on other vehicles, but if you can confirm this branch works it should be suitable for data capture on Thursday.

mojomex (commented):
I think there aren't any big issues with the implementation as-is. However, I think the use of std::variant makes parts of the code hard to read and maintain for the broader communitry, and I'd like it simplified. E.g. by using std::vector for the dynamic size arrays.

PR #421, opened 2026-03-19
feat(hesai): add CUDA-accelerated point cloud decoder

k1832:
## PR Type

- New Feature

## Description

Add a GPU-accelerated decode path for Hesai LiDAR sensors using CUDA. The feature is:

- **Compile-time opt-in**: Build with `-DBUILD_CUDA=ON`. When CUDA toolkit is not found, the build silently falls back to CPU-only.
- **Runtime opt-in**: Set `NEBULA_USE_CUDA=1` environment variable. When unset, the existing CPU path is used with zero overhead.

### What it does

- Processes an entire scan in a single batched CUDA kernel launch (`launch_decode_hesai_scan_batch`)
- Uses pre-computed angle lookup tables (azimuth/elevation) uploaded to GPU once at initialization
- Supports calibration-based and correction-based angle correctors
- Currently validated on OT128 (Pandar128E4X) sensor

### Files changed

| File | Change |
|------|--------|
| `hesai_cuda_kernels.cu` | New CUDA kernel for batched point cloud decoding |
| `hesai_cuda_decoder.hpp` | `HesaiScanDecoderCuda` RAII class — GPU buffer management, angle LUT, device memory |
| `hesai_decoder.hpp` | Integration: GPU scan buffer, flush, result conversion |
| `hesai_sensor.hpp` | Expose `max_scan_buffer_points()` for GPU buffer sizing |
| `angle_corrector_*.hpp` | Expose angle LUT data for GPU upload |
| `cuda_compat.hpp` | `NEBULA_HOST_DEVICE` / `NEBULA_DEVICE` macros for host/device code sharing |
| `nebula_hesai_decoders/CMakeLists.txt` | CUDA library target, toolkit detection |
| `nebula_hesai/CMakeLists.txt` | CUDA decoder test target |
| `hesai_cuda_decoder_test.cpp` | 5 GPU-vs-CPU equivalence tests |

### Known limitations

- GPU kernel does not set `return_type` field (always `return_id`)

### Performance

Measured using `profiling_runner.bash` (3 runs × 20s each, Pandar128E4X ~72k pts/scan, RTX 5080, CUDA 12.4):

```bash
./scripts/profiling_runner.bash cpu-baseline \
    --sensor-model Pandar128E4X --rosbag-path <ot128_rosbag> -n 3 -t 20

NEBULA_USE_CUDA=1 ./scripts/profiling_runner.bash gpu-cuda \
    --sensor-model Pandar128E4X --rosbag-path <ot128_rosbag> -n 3 -t 20

./scripts/plot_times.py cpu-baseline gpu-cuda --metrics decode
```

| Configuration | Median | P5 | P95 |
|---|---|---|---|
| CPU Baseline | 6.80 ms | 6.62 ms | 7.28 ms |
| GPU (this PR) | 2.48 ms | 2.41 ms | 12.77 ms |

The GPU results show a bimodal distribution: ~57% of scans decode in ~2.4 ms (~2.8x faster than CPU), while the remaining ~43% take ~10.5 ms. The slow path is dominated by the bulk D2H copy in `process_gpu_results()`, which copies the entire sparse output buffer back to host memory. The follow-up PR (zero-copy via `cuda_blackboard`) eliminates this D2H copy by keeping points on the GPU, which should remove the slow-path bottleneck.

### Build (with CUDA)

```bash
colcon build --packages-up-to nebula_hesai \
  --cmake-args -DBUILD_CUDA=ON -DBUILD_TESTING=ON
```

Requires NVIDIA CUDA Toolkit (tested with CUDA 12.x). If the toolkit is not found, the build succeeds but CUDA support is silently disabled.

### Running with CUDA enabled

The GPU decode path is gated by a runtime environment variable:

```bash
# Enable GPU decoding
export NEBULA_USE_CUDA=1

# Launch the driver node as usual — it will log "CUDA decoder initialized successfully" on startup
ros2 launch nebula nebula_launch.py sensor_model:=Pandar128E4X ...

# To disable (default), unset the variable
unset NEBULA_USE_CUDA
```

### Test

```bash
# Run all tests (132 existing + 5 new CUDA tests)
source install/setup.bash
colcon test --packages-select nebula_hesai --ctest-args -V

# Or run CUDA tests only
./build/nebula_hesai/hesai_cuda_decoder_test_main
```

### Test results

```
[==========] Running 5 tests from 1 test suite.
[ RUN      ] HesaiCudaDecoderTest.OT128_GpuVsCpuEquivalence
[       OK ] HesaiCudaDecoderTest.OT128_GpuVsCpuEquivalence (21778 ms)
[ RUN      ] HesaiCudaDecoderTest.OT128_GpuOutputNonEmpty
[       OK ] HesaiCudaDecoderTest.OT128_GpuOutputNonEmpty (388 ms)
[ RUN      ] HesaiCudaDecoderTest.OT128_GpuFieldValidity
[       OK ] HesaiCudaDecoderTest.OT128_GpuFieldValidity (378 ms)
[ RUN      ] HesaiCudaDecoderTest.OT128_BoundaryScanPointCounts
[       OK ] HesaiCudaDecoderTest.OT128_BoundaryScanPointCounts (369 ms)
[ RUN      ] HesaiCudaDecoderTest.OT128_IntensityExactMatch
[       OK ] HesaiCudaDecoderTest.OT128_IntensityExactMatch (17217 ms)
[  PASSED  ] 5 tests.

# Full suite
Summary: 137 tests, 0 errors, 0 failures, 0 skipped
```

## Remarks

- When CUDA is not compiled in (`BUILD_CUDA=OFF`), the 5 CUDA tests are compiled but skip at runtime via `GTEST_SKIP()`, so they do not break CPU-only CI.
- GPU uses CPU-authoritative scan cutting via `scan_state` flags. Point counts are identical between CPU and GPU (verified by `kMaxPointCountDiff = 0` in tests). Coordinate differences are sub-millimetre (< 0.1 mm) due to floating-point hardware rounding.

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/src/nebula_core/nebula_core_common/include/nebula_core_common/nebula_common.hpp
+++ b/src/nebula_core/nebula_core_common/include/nebula_core_common/nebula_common.hpp
@@ -15,6 +15,7 @@
 #ifndef NEBULA_COMMON_H
 #define NEBULA_COMMON_H
 
+#include <nebula_core_common/cuda_compat.hpp>
 #include <nebula_core_common/point_types.hpp>
 
 #include <boost/tokenizer.hpp>
@@ -507,13 +508,15 @@ PointCloud<PointXYZ> convert_point_xyzircaedt_to_point_xyz(
 /// @param radians
 /// @return degrees
 template <typename T>
-static inline std::enable_if_t<std::is_floating_point_v<T>, T> deg2rad(T degrees)
+NEBULA_HOST_DEVICE static inline constexpr std::enable_if_t<std::is_floating_point_v<T>, T> deg2rad(
+  T degrees)
 {
   return degrees * static_cast<T>(M_PI / 180.0);
 }
 
 template <typename T>
-static inline std::enable_if_t<std::is_integral_v<T>, double> deg2rad(T degrees)
+NEBULA_HOST_DEVICE static inline constexpr std::enable_if_t<std::is_integral_v<T>, double> deg2rad(
+  T degrees)
 {
   return deg2rad<double>(static_cast<double>(degrees));
 }
@@ -522,7 +525,8 @@ static inline std::enable_if_t<std::is_integral_v<T>, double> deg2rad(T degrees)
 /// @param radians
 /// @return degrees
 template <typename T>
-static inline std::enable_if_t<std::is_floating_point_v<T>, T> rad2deg(T radians)
+NEBULA_HOST_DEVICE static inline constexpr std::enable_if_t<std::is_floating_point_v<T>, T> rad2deg(
+  T radians)
 {
   return radians * static_cast<T>(180.0 / M_PI);
 }
--- a/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
+++ b/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
@@ -1,4 +1,4 @@
-// Copyright 2024 TIER IV, Inc.
+// Copyright 2026 TIER IV, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
+++ b/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
@@ -134,9 +134,9 @@ TEST(HesaiCudaDecoderTest, OT128_GpuVsCpuEquivalence)
 
     // For each GPU point, find nearest CPU point (brute-force — clouds are small enough)
     size_t matched = 0;
-    for (const auto & gp : gpu_cloud->points) {
+    for (const auto & gp : *gpu_cloud) {
       float best = std::numeric_limits<float>::max();
-      for (const auto & cp : cpu_cloud->points) {
+      for (const auto & cp : *cpu_cloud) {
         float d = sq_dist(gp, cp);
         if (d < best) best = d;
         if (d < tol_sq) break;  // early exit — found a match
@@ -173,7 +173,7 @@ TEST(HesaiCudaDecoderTest, OT128_GpuFieldValidity)
   ASSERT_GT(gpu_scans.size(), 0u);
 
   for (size_t i = 0; i < gpu_scans.size(); ++i) {
-    for (const auto & pt : gpu_scans[i].cloud->points) {
+    for (const auto & pt : *gpu_scans[i].cloud) {
       // Distance must be positive (all valid points)
       float dist = std::sqrt(pt.x * pt.x + pt.y * pt.y + pt.z * pt.z);
       EXPECT_GT(dist, 0.f) << "Scan " << i << ": zero-distance point found";
@@ -213,8 +213,8 @@ TEST(HesaiCudaDecoderTest, OT128_IntensityExactMatch)
   ASSERT_EQ(cpu_scans.size(), gpu_scans.size());
 
   for (size_t i = 0; i < cpu_scans.size(); ++i) {
-    const auto & cpu_pts = cpu_scans[i].cloud->points;
-    const auto & gpu_pts = gpu_scans[i].cloud->points;
+    const auto & cpu_pts = *cpu_scans[i].cloud;
+    const auto & gpu_pts = *gpu_scans[i].cloud;
 
     // For each GPU point, find the nearest CPU match and verify intensity is identical.
     // Only compare points that have a close spatial match (same physical point).
--- a/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
+++ b/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
@@ -34,6 +34,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdlib>
+#include <limits>
 #include <memory>
 #include <string>
 #include <vector>
--- a/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
+++ b/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
@@ -45,8 +45,17 @@ namespace nebula::test
 
 // OT128 config — matches TEST_CONFIGS[8] in hesai_ros_decoder_test_main.cpp
 static const nebula::ros::HesaiRosDecoderTestParams OT128_CONFIG = {
-  "Pandar128E4X", "LastStrongest", "Pandar128E4X.csv", "ot128/1730271167765338806",
-  "hesai", 0, 0.0, 0., 360., 0.3f, 300.f};
+  "Pandar128E4X",
+  "LastStrongest",
+  "Pandar128E4X.csv",
+  "ot128/1730271167765338806",
+  "hesai",
+  0,
+  0.0,
+  0.,
+  360.,
+  0.3f,
+  300.f};
 
 // Maximum allowed difference in point count between GPU and CPU per scan.
 // The GPU kernel assigns scan-boundary points differently than the CPU's ScanCutter,
@@ -81,13 +90,12 @@ static std::vector<DecodedScan> decode_bag(
   }
 
   rclcpp::NodeOptions options;
-  auto driver = std::make_shared<nebula::ros::HesaiRosDecoderTest>(
-    options, "cuda_test_node", params);
+  auto driver =
+    std::make_shared<nebula::ros::HesaiRosDecoderTest>(options, "cuda_test_node", params);
   EXPECT_EQ(driver->get_status(), nebula::Status::OK);
 
   std::vector<DecodedScan> scans;
-  auto cb = [&](uint64_t msg_ts, uint64_t /*scan_ts*/,
-                nebula::drivers::NebulaPointCloudPtr cloud) {
+  auto cb = [&](uint64_t msg_ts, uint64_t /*scan_ts*/, nebula::drivers::NebulaPointCloudPtr cloud) {
     if (cloud && !cloud->empty()) {
       // Deep-copy: the decoder reuses its internal buffer after the callback returns
       auto copy = std::make_shared<nebula::drivers::NebulaPointCloud>(*cloud);
@@ -99,8 +107,7 @@ static std::vector<DecodedScan> decode_bag(
 }
 
 /// Squared Euclidean distance between two points
-static float sq_dist(
-  const nebula::drivers::NebulaPoint & a, const nebula::drivers::NebulaPoint & b)
+static float sq_dist(const nebula::drivers::NebulaPoint & a, const nebula::drivers::NebulaPoint & b)
 {
   float dx = a.x - b.x;
   float dy = a.y - b.y;
@@ -247,8 +254,7 @@ TEST(HesaiCudaDecoderTest, OT128_IntensityExactMatch)
     // A small fraction (<1%) may differ at scan boundaries where the GPU kernel
     // selects a different return than the CPU's dual-return filtering.
     EXPECT_GT(checked, 0u) << "Scan " << i << ": no tight matches found";
-    double mismatch_ratio =
-      checked > 0 ? static_cast<double>(intensity_mismatch) / checked : 0.0;
+    double mismatch_ratio = checked > 0 ? static_cast<double>(intensity_mismatch) / checked : 0.0;
     EXPECT_LT(mismatch_ratio, 0.01)
       << "Scan " << i << ": " << intensity_mismatch << "/" << checked
       << " matched points have different intensity (" << (mismatch_ratio * 100) << "%)";
--- a/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
+++ b/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
@@ -16,14 +16,9 @@
 /// @brief GPU-vs-CPU equivalence tests for the Hesai CUDA decoder (OT128 / Pandar128E4X).
 ///
 /// Decodes the same rosbag with the CPU path (NEBULA_USE_CUDA unset) and the GPU path
-/// (NEBULA_USE_CUDA=1), then compares the resulting point clouds with tolerances.
-/// The GPU path differs at scan boundaries due to different overlap/FOV detection logic
-/// in the GPU kernel vs the CPU's ScanCutter.
-///
-/// Tolerances were derived from a single OT128 rosbag (ot128/1730271167765338806):
-///   - Scan 0: CPU=70275, GPU=72125 (diff=1850)
-///   - Scan 1: CPU=72240, GPU=72185 (diff=55)
-/// If additional rosbags show larger boundary diffs, tolerances may need adjustment.
+/// (NEBULA_USE_CUDA=1), then compares the resulting point clouds.
+/// The GPU path uses CPU-authoritative scan cutting (via scan_state flags), so scan
+/// boundaries are identical between CPU and GPU.
 
 #include "hesai_ros_decoder_test.hpp"
 
@@ -59,16 +54,14 @@ static const nebula::ros::HesaiRosDecoderTestParams OT128_CONFIG = {
   300.f};
 
 // Maximum allowed difference in point count between GPU and CPU per scan.
-// The GPU kernel assigns scan-boundary points differently than the CPU's ScanCutter,
-// causing up to ~2000 points to shift between adjacent scans.
-// Observed max diff: 1850 (scan 0 of test rosbag). Set to 2000 for ~8% headroom.
-static constexpr int kMaxPointCountDiff = 2000;
+// GPU uses CPU-authoritative scan_state, so scan cutting is identical.
+static constexpr int kMaxPointCountDiff = 0;
 // Coordinate tolerance (metres) for nearest-neighbour matching.
-// GPU uses pre-computed LUT with 0.01-degree resolution; small rounding differences expected.
-static constexpr float kXyzTolerance = 5e-3f;
+// GPU and CPU use the same angle LUT, but floating-point hardware differences
+// (e.g. FMA rounding on GPU vs CPU) cause sub-millimetre coordinate differences.
+static constexpr float kXyzTolerance = 1e-4f;  // 0.1 mm
 // Fraction of GPU points that must have a CPU match within tolerance.
-// ~2-3% of points are at scan boundaries and may not have an exact match.
-static constexpr double kMinMatchRatio = 0.97;
+static constexpr double kMinMatchRatio = 1.0;
 
 /// Decoded scan: message timestamp + point cloud
 struct DecodedScan
--- a/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
+++ b/src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
@@ -1,0 +1,278 @@
+// Copyright 2024 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// @file hesai_cuda_decoder_test.cpp
+/// @brief GPU-vs-CPU equivalence tests for the Hesai CUDA decoder (OT128 / Pandar128E4X).
+///
+/// Decodes the same rosbag with the CPU path (NEBULA_USE_CUDA unset) and the GPU path
+/// (NEBULA_USE_CUDA=1), then compares the resulting point clouds with tolerances.
+/// The GPU path differs at scan boundaries due to different overlap/FOV detection logic
+/// in the GPU kernel vs the CPU's ScanCutter.
+///
+/// Tolerances were derived from a single OT128 rosbag (ot128/1730271167765338806):
+///   - Scan 0: CPU=70275, GPU=72125 (diff=1850)
+///   - Scan 1: CPU=72240, GPU=72185 (diff=55)
+/// If additional rosbags show larger boundary diffs, tolerances may need adjustment.
+
+#include "hesai_ros_decoder_test.hpp"
+
+#include <rclcpp/rclcpp.hpp>
+
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <cmath>
+#include <cstdlib>
+#include <memory>
+#include <string>
+#include <vector>
+
+#ifdef NEBULA_CUDA_ENABLED
+
+namespace nebula::test
+{
+
+// OT128 config — matches TEST_CONFIGS[8] in hesai_ros_decoder_test_main.cpp
+static const nebula::ros::HesaiRosDecoderTestParams OT128_CONFIG = {
+  "Pandar128E4X", "LastStrongest", "Pandar128E4X.csv", "ot128/1730271167765338806",
+  "hesai", 0, 0.0, 0., 360., 0.3f, 300.f};
+
+// Maximum allowed difference in point count between GPU and CPU per scan.
+// The GPU kernel assigns scan-boundary points differently than the CPU's ScanCutter,
+// causing up to ~2000 points to shift between adjacent scans.
+// Observed max diff: 1850 (scan 0 of test rosbag). Set to 2000 for ~8% headroom.
+static constexpr int kMaxPointCountDiff = 2000;
+// Coordinate tolerance (metres) for nearest-neighbour matching.
+// GPU uses pre-computed LUT with 0.01-degree resolution; small rounding differences expected.
+static constexpr float kXyzTolerance = 5e-3f;
+// Fraction of GPU points that must have a CPU match within tolerance.
+// ~2-3% of points are at scan boundaries and may not have an exact match.
+static constexpr double kMinMatchRatio = 0.97;
+
+/// Decoded scan: message timestamp + point cloud
+struct DecodedScan
+{
+  uint64_t msg_timestamp;
+  nebula::drivers::NebulaPointCloudPtr cloud;
+};
+
+/// Decode a rosbag with the given CUDA env setting.
+/// When @p use_cuda is true, sets NEBULA_USE_CUDA=1; otherwise unsets it.
+/// Returns one DecodedScan per completed scan.
+static std::vector<DecodedScan> decode_bag(
+  const nebula::ros::HesaiRosDecoderTestParams & params, bool use_cuda)
+{
+  // Toggle GPU/CPU path via environment variable
+  if (use_cuda) {
+    setenv("NEBULA_USE_CUDA", "1", 1);
+  } else {
+    unsetenv("NEBULA_USE_CUDA");
+  }
+
+  rclcpp::NodeOptions options;
+  auto driver = std::make_shared<nebula::ros::HesaiRosDecoderTest>(
+    options, "cuda_test_node", params);
+  EXPECT_EQ(driver->get_status(), nebula::Status::OK);
+
+  std::vector<DecodedScan> scans;
+  auto cb = [&](uint64_t msg_ts, uint64_t /*scan_ts*/,
+                nebula::drivers::NebulaPointCloudPtr cloud) {
+    if (cloud && !cloud->empty()) {
+      // Deep-copy: the decoder reuses its internal buffer after the callback returns
+      auto copy = std::make_shared<nebula::drivers::NebulaPointCloud>(*cloud);
+      scans.push_back({msg_ts, copy});
+    }
+  };
+  driver->read_bag(cb);
+  return scans;
+}
+
+/// Squared Euclidean distance between two points
+static float sq_dist(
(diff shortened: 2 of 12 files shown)


mojomex (commented):
Thank you for the awesome PR!

I think the core of it is already good, and it's quite readable and cleanly implemented. I've left some comments here and there about what would make our maintenance life easier moving forward.

About the scan cutting differences, I think you could fairly easily integrate the CPU scan cutting state (which is still available, even when CUDA is enabled in your implementation) to make CUDA behave exactly like the CPU version.

I haven't yet reviewed every detail completely, but there's enough feedback for now I think.

Also, I'd really appreciate if you could post some performance metrics, e.g. using [Nebula's benchmark scripts](https://tier4.github.io/nebula/contributing/benchmarking/#running-benchmarks).

src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+    size_t matched = 0;
+    for (const auto & gp : *gpu_cloud) {
+      float best = std::numeric_limits<float>::max();
+      for (const auto & cp : *cpu_cloud) {
+        float d = sq_dist(gp, cp);
+        if (d < best) best = d;
+        if (d < tol_sq) break;  // early exit — found a match
+      }
+      if (best < tol_sq) ++matched;
+    }
```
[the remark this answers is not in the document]
k1832: Agreed. `kMaxPointCountDiff=0` plus the existing field validity / boundary count tests should cover this.

src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+// Maximum allowed difference in point count between GPU and CPU per scan.
+// The GPU kernel assigns scan-boundary points differently than the CPU's ScanCutter,
+// causing up to ~2000 points to shift between adjacent scans.
+// Observed max diff: 1850 (scan 0 of test rosbag). Set to 2000 for ~8% headroom.
+static constexpr int kMaxPointCountDiff = 2000;
+// Coordinate tolerance (metres) for nearest-neighbour matching.
+// GPU uses pre-computed LUT with 0.01-degree resolution; small rounding differences expected.
+static constexpr float kXyzTolerance = 5e-3f;
+// Fraction of GPU points that must have a CPU match within tolerance.
+// ~2-3% of points are at scan boundaries and may not have an exact match.
```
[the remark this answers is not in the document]
k1832: Thank you for the review!

All inline comments have been addressed.

Re performance: I added a comparison to the PR description (CPU 6.08ms vs GPU 0.31ms, ~20x speedup on RTX 5080). The numbers were collected using a non-standard profiling setup. Let me look into reproducing them with `profiling_runner.bash` / `plot_times.py` and I will follow up with that.

mojomex (commented):
@k1832 Thank you for addressing the changes!
It seems like they're not pushed yet though, please check.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
```diff
@@ -342,8 +594,14 @@ class HesaiDecoder : public HesaiScanDecoder
         SensorT::peak_resolution_mdeg.azimuth, SensorT::packet_t::n_channels,
         logger_->child("Downsample Mask"), true, sensor_.get_dither_transform());
     }
+
+#ifdef NEBULA_CUDA_ENABLED
+    initialize_cuda();
+#endif
   }
 
+  // unique_ptr<HesaiScanDecoderCuda> handles CUDA resource cleanup automatically
```
mojomex: I think this comment is not needed.

src/nebula_hesai/nebula_hesai_decoders/CMakeLists.txt
```diff
@@ -20,6 +43,52 @@ target_link_libraries(
...
+  target_link_libraries(nebula_hesai_decoders_cuda
+    PRIVATE CUDA::cudart_static)
+
+  target_compile_definitions(nebula_hesai_decoders_cuda PUBLIC NEBULA_CUDA_ENABLED)
+  target_compile_definitions(nebula_hesai_decoders PUBLIC NEBULA_CUDA_ENABLED)
+
+  # Add CUDA include directories to main library for header resolution
+  target_include_directories(nebula_hesai_decoders
+    PUBLIC ${CUDAToolkit_INCLUDE_DIRS})
+
```
mojomex: Unsure whether CUDA is some special case, but usually doing `target_link_libaries` to the used CUDA targets should be enough, making this include_directories unnecessary.

src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+#endif  // NEBULA_CUDA_ENABLED
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  int result = RUN_ALL_TESTS();
+  // Restore env
+  unsetenv("NEBULA_USE_CUDA");
```
mojomex: (nit) getting the actual value of NEBULA_USE_CUDA on startup, then restoring the original would be more robust. Should never be set on startup anyway though, so feel free to ignore.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    const size_t lut_size = angle_lut.size() * sizeof(CudaAngleCorrectionData);
+    cudaError_t err = cudaMalloc(&d_angle_lut_, lut_size);
+    if (err != cudaSuccess) {
+      fprintf(stderr, "CUDA: Failed to allocate angle LUT: %s\n", cudaGetErrorString(err));
+      return false;
+    }
+
+    err = cudaMemcpy(d_angle_lut_, angle_lut.data(), lut_size, cudaMemcpyHostToDevice);
```
manato:
```suggestion
    err = cudaMemcpyAsync(d_angle_lut_, angle_lut.data(), lut_size, cudaMemcpyHostToDevice, stream_);
```

`cudaMemcpy` triggers implicit synchronization, which may block other CUDA operations running on other threads in the same process (In the case that we use a ROS2 composable container that consists of this implementation and other CUDA implemented nodes, this `cudaMemcpy` may block other CUDA operations in other nodes).

Even though this function seems to be called once during start-up, I would suggest using the async version to avoid unexpected blocking.

manato (commented):
@k1832
Thank you very much for the PR!

Overall, it is nicely looking and well organized!
I left some comments that I noticed. I'd appreciate it if you could consider them :pray:

In terms of speed up, I guess measurement result using rosbag and the actual sensor would be slightly different because the case of using the actual sensor does pipelined-processing of decoding and receiving, while the case of using rosbag proceeds whole 1 scan packet (i.e., accumulated packets) at once.
By any chance, do you perhaps have any measurement data using the actual sensors?

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,383 @@
...
+      d_angle_lut_, config, d_points_, d_count_, n_azimuths_, n_entries, stream_);
+
+    if (!kernel_ok) {
+      return 0;
+    }
+
+    cudaStreamSynchronize(stream_);
+
+    uint32_t valid_point_count = 0;
+    cudaMemcpy(&valid_point_count, d_count_, sizeof(uint32_t), cudaMemcpyDeviceToHost);
```
manato:
```suggestion
    uint32_t valid_point_count = 0;
    cudaMemcpyAsync(&valid_point_count, d_count_, sizeof(uint32_t), cudaMemcpyDeviceToHost, stream_);

    cudaStreamSynchronize(stream_);
```

CUDA operations called on the same stream are guaranteed to be executed in the issued order. Hence, I believe you can skip `cudaStreamSynchronize` before copying back `valid_point_count` (and would prefer to avoid `cudaMemcpy` for the same reason discussed before).

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
```diff
@@ -96,6 +103,21 @@ class HesaiDecoder : public HesaiScanDecoder
...
+  std::unique_ptr<cuda::HesaiScanDecoderCuda> cuda_scan_decoder_;
+
+  /// @brief Number of azimuth divisions for angle lookup table (LUT resolution)
+  static constexpr uint32_t cuda_n_azimuths_ = 36000;  // 0.01 degree resolution
+  /// @brief Sensor's native azimuth range (max_azimuth = 360 * degree_subdivisions)
+  static constexpr uint32_t sensor_max_azimuth_ = 360 * SensorT::packet_t::degree_subdivisions;
+  /// @brief Scale factor from sensor native azimuth to LUT index
+  static constexpr uint32_t azimuth_scale_ = sensor_max_azimuth_ / cuda_n_azimuths_;
+
+  static constexpr uint32_t MAX_PACKETS_PER_SCAN = 4000;
```
manato: [ask] Does the value 4000 come from any specification? I'm just curious if this amount can be reasonably enough for all supported Hesai sensors

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,383 @@
...
+
+    if (valid_point_count > 0) {
+      // Copy sparse buffer to host
+      const uint32_t copy_size =
+        std::min(sparse_buffer_size, static_cast<uint32_t>(host_point_buffer_.size()));
+      cudaMemcpy(
+        host_point_buffer_.data(), d_points_, copy_size * sizeof(CudaNebulaPoint),
+        cudaMemcpyDeviceToHost);
```
manato:
```suggestion
cudaMemcpyAsync(
        host_point_buffer_.data(), d_points_, copy_size * sizeof(CudaNebulaPoint),
        cudaMemcpyDeviceToHost, stream_);

      cudaStreamSynchronize(stream_);
```

src/nebula_hesai/nebula_hesai_decoders/src/cuda/hesai_cuda_kernels.cu
```diff
@@ -0,0 +1,164 @@
...
+
+  if (n_returns == 2) {
+    const uint32_t last_idx = group_base + 1;
+    const uint16_t last_raw_distance = d_distances_batch[last_idx];
+    const uint8_t last_reflectivity = d_reflectivities_batch[last_idx];
+
+    if (raw_distance == last_raw_distance && reflectivity == last_reflectivity) return true;
+
+    const float last_distance = static_cast<float>(last_raw_distance) * dis_unit;
+    if (fabsf(distance - last_distance) < threshold) return true;
```
manato:
```suggestion
    return (fabsf(distance - last_distance) < threshold);
```

I believe just returning the condition that is currently evaluated in `if` can reduce the opportunity of "warp divergence".

src/nebula_hesai/nebula_hesai_decoders/src/cuda/hesai_cuda_kernels.cu
```diff
@@ -0,0 +1,164 @@
...
+    for (uint32_t other_ret = 0; other_ret < n_returns; ++other_ret) {
+      if (other_ret == return_id) continue;
+
+      const uint16_t other_raw_distance = d_distances_batch[group_base + other_ret];
+      const uint8_t other_reflectivity = d_reflectivities_batch[group_base + other_ret];
+
+      if (raw_distance == other_raw_distance && reflectivity == other_reflectivity) return true;
+
+      const float other_distance = static_cast<float>(other_raw_distance) * dis_unit;
+      if (fabsf(distance - other_distance) < threshold) return true;
```
manato:
```suggestion
      return (fabsf(distance - other_distance) < threshold);
```

src/nebula_hesai/nebula_hesai_decoders/src/cuda/hesai_cuda_kernels.cu
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+  return false;
```
manato:
```suggestion

```

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,383 @@
...
+      cudaMemcpyHostToDevice, stream_);
+    cudaMemcpyAsync(
+      d_raw_azimuths_, h_raw_azimuths_staging_, n_entries * sizeof(uint32_t),
+      cudaMemcpyHostToDevice, stream_);
+    cudaMemcpyAsync(
+      d_n_returns_, h_n_returns_staging_, n_entries * sizeof(uint32_t), cudaMemcpyHostToDevice,
+      stream_);
+    cudaMemcpyAsync(
+      d_scan_flags_, h_scan_flags_staging_, n_entries * n_channels_ * sizeof(uint8_t),
+      cudaMemcpyHostToDevice, stream_);
```
manato: As mentioned in the [official document](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g85073372f776b4c4d5f89f7124b7bf79), these API call may return codes from previous, asynchronous launches. Hence, it would be appreciated if you consider checking returned value, similar to the `check()` in the constructor.

Besides, I found that these copies take almost 5x execution time compared with `decode_hesai_scan_batch_kernel` mainly because these copies are executed on the same stream sequentially. Because the number of copies is fixed to 5, and they are independent from each other, it might be better to consider creating dedicated 5 CUDA streams for data copy.

Screenshot from 2026-04-03 22-36-08

mojomex (changes requested):
Getting closer! I left some CUDA performance and architecture comments, please check.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
```diff
@@ -96,6 +103,21 @@ class HesaiDecoder : public HesaiScanDecoder
 
   std::array<DecodeFrame, 2> frame_buffers_{initialize_frame(), initialize_frame()};
 
+#ifdef NEBULA_CUDA_ENABLED
```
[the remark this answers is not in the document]
k1832: Agreed this would be cleaner. This is a larger refactor — will address in a follow-up.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
k1832: Removed.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/angle_corrector_correction_based.hpp
```diff
@@ -25,8 +25,11 @@
 #include <cassert>
 #include <cstdint>
 #include <memory>
+#include <sstream>
 #include <stdexcept>
 #include <string>
+#include <tuple>
+#include <vector>
```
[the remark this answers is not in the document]
k1832: The two functions are removed. Also removed the now-unused `#include <tuple>`.

src/nebula_hesai/nebula_hesai_decoders/CMakeLists.txt
k1832: Removed the explicit `target_include_directories`. `CUDA::cudart` propagates includes.

src/nebula_hesai/nebula_hesai/tests/hesai_cuda_decoder_test.cpp
k1832: Done — now saves and restores the original value.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
k1832: Done — using `cudaMemcpyAsync` on the dedicated stream + `cudaStreamSynchronize`.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
k1832: Done — count readback and result copy both use `cudaMemcpyAsync` on `stream_` with a single `cudaStreamSynchronize` at the end.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
k1832: Not from a spec — it is a conservative upper bound. OT128 in LastStrongest mode uses ~280 entries/scan (1 block group per packet, ~72k points / 256 points per entry). 4000 provides ~14x headroom to cover other Hesai sensors and single-return modes.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
k1832: Done — same fix as the count readback.

src/nebula_hesai/nebula_hesai_decoders/src/cuda/hesai_cuda_kernels.cu
k1832: Applied for the n_returns==2 path. Kept the original `if` + `return true` in the n_returns>2 loop since returning the condition there would exit early without checking remaining returns.

src/nebula_hesai/nebula_hesai_decoders/src/cuda/hesai_cuda_kernels.cu
k1832: This is inside a loop over all other returns — replacing `if (cond) return true` with `return (cond)` would exit on the first non-matching return without checking the rest. Kept the original logic for correctness with n_returns > 2.

src/nebula_hesai/nebula_hesai_decoders/src/cuda/hesai_cuda_kernels.cu
k1832: Removed (clang-format).

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
k1832: Good points. Switched all copies to `cudaMemcpyAsync` on the dedicated stream. Multiple streams for parallel H2D is a good optimization idea — will consider as a follow-up.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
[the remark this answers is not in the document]
k1832: Fixed — all `cudaMemcpy` replaced with `cudaMemcpyAsync` on `stream_`, and sync uses `cudaStreamSynchronize(stream_)` only.

k1832:
@drwnz (CC: @mojomex )

> my main concern is about the actual latency with a real sensor.

Good point. Here's my analysis of the latency difference:

If my understanding is correct, both CPU and GPU paths publish the full scan's point cloud at once when the scan boundary is detected, not per-packet. The difference is how much work remains at that moment:

- **CPU**: points decoded incrementally as packets arrived. At boundary, just publish. Near-instant.
- **GPU**: packets only staged to host memory. At boundary, full flush (H2D + kernel + D2H) before publishing.

So the GPU path adds flush latency at the scan boundary. The profiling results show this flush takes 2.4ms on some scans and 10.5ms on others (bimodal pattern), where the slow path is dominated by the bulk D2H sparse buffer copy. This bimodal pattern is likely an artifact of the short test rosbag being looped. The follow-up PR (zero-copy via cuda_blackboard) eliminates this D2H copy entirely, which should remove the slow path and make the distribution uniform.

The primary benefit of PR1 is **CPU offload** during the scan period. This is fully opt-in — build with `-DBUILD_CUDA=OFF` and no CUDA code is compiled at all. Even with CUDA compiled in, omitting `export NEBULA_USE_CUDA=1` skips GPU initialization, and the only runtime cost is a single null pointer check per packet.

Again, it's a great discussion point. I need your team's perspective and let's keep this discussion going here.

src/nebula_core/nebula_core_common/include/nebula_core_common/nebula_common.hpp
```diff
@@ -15,6 +15,7 @@
 #ifndef NEBULA_COMMON_H
 #define NEBULA_COMMON_H
 
+#include <nebula_core_common/cuda_compat.hpp>
```
nishikawa-masaki, line 18: The coding style guide recommends naming with regular words, not with uncommon abbreviations.
"cuda_compatibility.hpp" is better, isn't it?

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,385 @@
...
+public:
+  /// @brief Construct and allocate all GPU resources.
+  /// @param n_channels Number of LiDAR channels (e.g. 128 for OT128)
+  /// @param max_returns Maximum returns per block (e.g. 2 for dual return)
+  /// @param max_packets Maximum packets per scan
+  /// @param max_output_points Upper bound on output points (for sparse buffer sizing)
+  /// @param n_azimuths Number of azimuth divisions in the angle LUT
+  /// @param azimuth_scale Scale factor from sensor azimuth to LUT index
+  /// @throws std::runtime_error if any CUDA allocation fails
+  HesaiScanDecoderCuda(
```
nishikawa-masaki, line 100: Autoware, ROS, Google Coding Style Guide allows writing definition (code) of a class function in a header file only when the function is short (less than or equal to 10 lines).
If the function is more complex, only its declaration should be written in the header file and its definition (actual code) should reside in .cpp file.
https://google.github.io/styleguide/cppguide.html#Defining_Functions_in_Header_Files

Personally, I don't like code lines in header files, especially if there is only a '.hpp' file with code lines and there is no related '*.cpp' file.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,385 @@
...
+  // Pinned host staging buffers
+  uint16_t * h_distances_staging_ = nullptr;
+  uint8_t * h_reflectivities_staging_ = nullptr;
+  uint32_t * h_raw_azimuths_staging_ = nullptr;
+  uint32_t * h_n_returns_staging_ = nullptr;
+  uint8_t * h_scan_flags_staging_ = nullptr;
+  uint64_t * h_packet_timestamps_staging_ = nullptr;
+
+  // Angle correction LUT on device
+  CudaAngleCorrectionData * d_angle_lut_ = nullptr;
```
nishikawa-masaki, line 379: It would be better if these pointers to resources were std::unique_ptr.
With std::unique_ptr, you can avoid the following problems.
* Resource leaking in the constructor
  * When the constructor throws an exception, the destructor won't be executed.
* Copying and moving of the pointers.
* Forgetting to free resources in the destructor (in future).

However, you need to give them special deleters.

nishikawa-masaki:
Since Autoware Jazzy, private (non-exported) header files are expected to be located in src or src/include.
Please read the following URL.
https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ros-nodes/directory-structure/

Already existing headers may need another PR.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,385 @@
...
+  : n_channels_(n_channels),
+    max_returns_(max_returns),
+    max_packets_(max_packets),
+    max_output_points_(max_output_points),
+    n_azimuths_(n_azimuths),
+    azimuth_scale_(azimuth_scale)
+  {
+    auto check = [](cudaError_t err, const char * name) {
+      if (err != cudaSuccess) {
+        throw std::runtime_error(
```
nishikawa-masaki, line 112: The destructor won't be executed if an exception is thrown in the constructor (because the object has not been fully constructed yet).
It can lead to resource leaking.
Use of std::unique_ptr can be a solution.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,385 @@
...
+  HesaiScanDecoderCuda(HesaiScanDecoderCuda &&) = delete;
+  HesaiScanDecoderCuda & operator=(HesaiScanDecoderCuda &&) = delete;
+
+  /// @brief Upload angle correction lookup table to GPU
+  bool upload_angle_corrections(
+    const std::vector<CudaAngleCorrectionData> & angle_lut, uint32_t n_azimuths,
+    uint32_t n_channels)
+  {
+    if (angle_lut.size() != static_cast<size_t>(n_azimuths) * n_channels) {
+      fprintf(
```
nishikawa-masaki, line 211: Can fprintf() be used in production code?

nishikawa-masaki (changes requested):
I haven't finished reviewing, but I submit my comments as of now.

nishikawa-masaki:
FYI
I asked the following question to Github CoPilot.
This is just a hint and can be addressed in another PR.
Q.
Is there any possible option for improving performance of hesai_cuda_kernels.cu?
A. (Extracted only the answer with the highest impact)
Highest Impact
Compact output on GPU and copy only valid points
Current behavior writes point data to sparse index by thread id at hesai_cuda_kernels.cu:125, still does an atomic count at hesai_cuda_kernels.cu:138, then zeros/copies/scans the whole sparse buffer on host at hesai_cuda_decoder.hpp:312, hesai_cuda_decoder.hpp:331, hesai_decoder.hpp:227.
Better approach: use atomic result as output index, write contiguous output, copy only valid_count points, iterate only valid_count on CPU.
This removes three large costs at once: full buffer memset, full sparse device-to-host copy, and sparse host scan.
In pipelines like this, this is often the largest end-to-end speedup.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
```diff
@@ -0,0 +1,385 @@
...
+        throw std::runtime_error(
+          std::string("CUDA allocation failed for ") + name + ": " + cudaGetErrorString(err));
+      }
+    };
+
+    check(cudaStreamCreate(&stream_), "CUDA stream");
+
+    // Output buffers
+    check(
+      cudaMalloc(&d_points_, max_output_points_ * sizeof(CudaNebulaPoint)), "output points buffer");
```
nishikawa-masaki, line 121: cudaMallocAsync() may be preferable to cudaMalloc().
* cudaMalloc() is executed in the default stream and stops all other streams.
  * cudaMallocAsync() is executed in the specified stream without any interference to other streams.
* cudaMallocAsync() uses internal memory pool.

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
mojomex: Very good point, I would also prefer that. One possible implementation is in [autoware_cuda_utils](https://github.com/autowarefoundation/autoware_universe/blob/main/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_unique_ptr.hpp).

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/cuda/hesai_cuda_decoder.hpp
mojomex: Good point, it shouldn't. A `nebula::util::Logger` should be passed in instead. It is an abstract interface that allows logging to ROS2/console without deep library code like this needing to depend on ROS2.

PR #429, opened 2026-03-31
test: cover nebula_velodyne_common calibration

mojomex:
## Summary
- add focused unit tests for `nebula_velodyne_common`
- cover return-mode parsing, status streaming, sensor configuration streaming, and calibration YAML read-write paths
- fix the package public link interface so `nebula_core_common` headers propagate correctly

## Verification
- `source /opt/ros/humble/setup.bash && colcon build --packages-select nebula_velodyne_common --symlink-install --mixin debug compile-commands coverage-gcc`
- `source /opt/ros/humble/setup.bash && colcon test --packages-select nebula_velodyne_common --event-handlers console_direct+`


drwnz (changes requested):
Thanks, looks great! The only ones that don't get tested here are failures in reading or writing the calibration file (bad file path, no file exists, malformed yaml). I don't know what the current actual behaviour is in this case, but a silent fail here is potentially a bad day for integration teams.

PR #438, opened 2026-04-19
fix(seyond): use correct rosdep key for nlohmann_json

otenim:
## Summary
- `nebula_seyond_common/package.xml` declared `<depend>nlohmann_json</depend>`, which is not a valid rosdep key. `rosdep install --from-paths . --ignore-src -y` failed with:
  ```
  ERROR: the following packages/stacks could not have their rosdep keys resolved to system dependencies:
  nebula_seyond_common: Cannot locate rosdep definition for [nlohmann_json]
  ```
- Replace the invalid key with the correct one, `nlohmann-json-dev`.

## Tests
- [x] `rosdep install --from-paths . --ignore-src -y` no longer reports the unresolved key
- [x] `colcon build --packages-up-to nebula` still succeeds

### DIFF
--- a/src/nebula_seyond/nebula_seyond/scripts/replay_seyond_pcap.py
+++ b/src/nebula_seyond/nebula_seyond/scripts/replay_seyond_pcap.py
@@ -1,28 +1,30 @@
 #!/usr/bin/env python3
 
 import argparse
+import os
 import socket
 import struct
-import time
 import sys
-import os
+import time
+
 
 class FastPcapReader:
     """
     A minimal, high-performance PCAP reader in pure Python.
     """
+
     def __init__(self, filename):
         self.f = open(filename, "rb")
         global_header = self.f.read(24)
         if len(global_header) < 24:
             raise ValueError("Invalid PCAP file")
-        
+
         # Check magic number for endianness
         magic = struct.unpack("I", global_header[0:4])[0]
-        if magic == 0xA1B2C3D4: # Native endian
-            self.fmt_header = "IIII" 
-            self.fmt_global = "IHHIIII" # magic, major, minor, gmt, sigfigs, snaplen, network
-        elif magic == 0xD4C3B2A1: # Swapped endian
+        if magic == 0xA1B2C3D4:  # Native endian
+            self.fmt_header = "IIII"
+            self.fmt_global = "IHHIIII"  # magic, major, minor, gmt, sigfigs, snaplen, network
+        elif magic == 0xD4C3B2A1:  # Swapped endian
             self.fmt_header = "<IIII"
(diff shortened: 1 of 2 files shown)


otenim:
About the pre-commit.ci error:

i guess its not caused by this patch, so im leaving it as is.

drwnz (approved):
(see later review)

drwnz (changes requested):
Apologies, I previously approved this but actually, it's not the right fix.
The correct fix is added as a suggestion. This package still depends on `nlohmann_json` but this change stops `nebula_seyond_common` from correctly declaring its own direct dependency for isolated rosdep/install/release use. The fix should be to use the correct rosdep key, `nlohmann-json-dev,` rather than deleting the dependency altogether.

Also, please sign your commits. Thanks!

otenim:
@drwnz 
Thanks for reviewing!!

> This package still depends on nlohmann_json but this change stops nebula_seyond_common from correctly declaring its own direct dependency for isolated rosdep/install/release use. 

Its right. I just replaced `nlohmann_json` with `nlohmann-json-dev` on `src/nebula_seyond/nebula_seyond_common/package.xml` and reverted the other files back to the original state (only the rosdep key change remains).

> please sign your commits. Thanks!
Added my sign on the commits.
I'll keep this convention in mind from now on.


On pre-commit & spell-check errors:

The errors seem to be caused by a python script that is not associated with this pr.
I would make another pr to fix those errors if ok.

drwnz:
> @drwnz Thanks for reviewing!!
> 
> > This package still depends on nlohmann_json but this change stops nebula_seyond_common from correctly declaring its own direct dependency for isolated rosdep/install/release use.
> 
> Its right. I just replaced `nlohmann_json` with `nlohmann-json-dev` on `src/nebula_seyond/nebula_seyond_common/package.xml` and reverted the other files back to the original state (only the rosdep key change remains).
> 
> > please sign your commits. Thanks!
> > Added my sign on the commits.
> > I'll keep this convention in mind from now on.
> 
> On pre-commit & spell-check errors:
> 
> The errors seem to be caused by a python script that is not associated with this pr. I would make another pr to fix those errors if ok.

Thanks, it seems you've commited pre-commit changes to other files. I'd leave them out of this PR - actually that python script is just for local testing and I was going to remove it later anyway.

otenim:
@drwnz 

Sorry for making you confused...

image 

Apparently it was an autofix push by pre-commit.ci (I didn't touch any other files than package.xml)

So, if i revert the autofix, i think it will just get re-applied.

Should we leave it as is?

drwnz (approved):
LGTM, thanks!

PR #444, opened 2026-05-18
fix: update port configurations for ARS548

Naophis:
## PR Type
- Improvement

## Related Links
https://star4.slack.com/archives/C076E8EBJTG/p1778655511314399

## Description

Since the port numbers used in ARS548, including `data_port`, are ephemeral ports, communication will fail if they are already occupied on the Host PC side.

Although users can flexibly change these values because they are parameterized, this PR updates the default values to prevent such conflicts and issues.

- [ ] Assign PR to reviewer

## CI Checks

### DIFF
--- a/src/nebula_continental/nebula_continental/config/ARS548.param.yaml
+++ b/src/nebula_continental/nebula_continental/config/ARS548.param.yaml
@@ -2,15 +2,15 @@
   ros__parameters:
     host_ip: 10.13.1.166
     sensor_ip: 10.13.1.113
-    data_port: 42102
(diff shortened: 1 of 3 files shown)


Naophis:
Unfortunately, data_port must be fixed port by devices (ARS548).
It looks like we'll handle this by adding an ephemeral port exclusion setting. 

### Example Configuration:

write in `/etc/sysctl.d/99-local-port.conf` 
```
net.ipv4.ip_local_reserved_ports = 42101,42102,42401
```
and execute `sudo sysctl --system`

PR #451, opened 2026-06-11
test: runtime common return mode

drwnz:
# PR 1 of 8: Cover Generic Return Mode Parsing
Tracked by https://github.com/tier4/nebula/issues/452

Branch: `test/runtime-common-return-mode`
Base: `main`
Type: Improvement

## Description

Adds a focused regression test for the generic `ReturnMode` string contract in
`nebula_core_common`.

This PR intentionally does not change runtime behavior. It documents the current
contract that only generic return mode names are parsed by
`return_mode_from_string()`, while vendor-specific combinations such as
`LastStrongest` and `FirstStrongest` remain vendor-owned and return `UNKNOWN` at
the common layer.

## Review Procedure

Review only:

- `src/nebula_core/nebula_core_common/CMakeLists.txt`
- `src/nebula_core/nebula_core_common/test/test_core_types.cpp`

Confirm that the new test target is minimal and does not introduce vendor or
runtime dependencies.

## Covered Validation

- `colcon build --packages-select nebula_core_common --event-handlers console_direct+ --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo`

## Notes

This PR is a test-only base for the stack. It does not add vendor-neutral runtime
interfaces or vendor driver support.


src/nebula_core/nebula_core_common/test/test_core_types.cpp
```diff
@@ -0,0 +1,53 @@
...
+using nebula::drivers::return_mode_from_string;
+using nebula::drivers::ReturnMode;
+
+TEST(TestCoreTypes, ReturnModeRoundTrip)
+{
+  const std::vector<std::pair<ReturnMode, std::string>> cases = {
+    {ReturnMode::SINGLE_FIRST, "SingleFirst"},
+    {ReturnMode::SINGLE_STRONGEST, "SingleStrongest"},
+    {ReturnMode::SINGLE_LAST, "SingleLast"},
+    {ReturnMode::DUAL_ONLY, "Dual"},
```
mojomex: `dual` is not a well-defined return mode, and does not correspond to a specific dual return setting. Currently, conventions seem to align across vendors, but I'd like it to be represented as a specific, e.g. `LastStrongest` option internally, or for it to be deprecated.

src/nebula_core/nebula_core_common/test/test_core_types.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <sstream>
+#include <string>
+#include <utility>
+#include <vector>
+
+using nebula::drivers::return_mode_from_string;
+using nebula::drivers::ReturnMode;
+
+TEST(TestCoreTypes, ReturnModeRoundTrip)
```
mojomex: Cases are not exhaustive, please add all return modes we want to support.

src/nebula_core/nebula_core_common/test/test_core_types.cpp
```diff
@@ -0,0 +1,53 @@
...
+    std::stringstream stream;
+    stream << test_case.first;
+    EXPECT_EQ(stream.str(), test_case.second);
+    EXPECT_EQ(return_mode_from_string(test_case.second), test_case.first);
+  }
+
+  EXPECT_EQ(return_mode_from_string("DualOnly"), ReturnMode::UNKNOWN);
+  EXPECT_EQ(return_mode_from_string("LastStrongest"), ReturnMode::UNKNOWN);
+  EXPECT_EQ(return_mode_from_string("FirstStrongest"), ReturnMode::UNKNOWN);
+  EXPECT_EQ(return_mode_from_string("Unknown"), ReturnMode::UNKNOWN);
```
mojomex: Why are these `UNKNOWN`? Shouldn't they be valid?

mojomex (changes requested):
Please have a look!

src/nebula_core/nebula_core_common/test/test_core_types.cpp
drwnz: Addressed in a39d7cde9fb3ca079300f9ce73a1ad7257def50d

src/nebula_core/nebula_core_common/test/test_core_types.cpp
drwnz: Addressed in a39d7cde9fb3ca079300f9ce73a1ad7257def50d

src/nebula_core/nebula_core_common/test/test_core_types.cpp
drwnz: I've done a bit of a refactor of return modes because previously types and modes were being confused.
This also leads to some vendor-specific edits and documentation changes. Please have a look and check that you agree - biggest design decision is how to consider the difference between "StrongestLast" and "LastStrongest". I chose to ignore the UDP field order in the concept of return mode, but kept compatibility with the old canonical strings.

PR #460, opened 2026-07-03
feat(nebula_core, nebula_hesai): register hesai LiDAR packet threads with cie_thread_configurator

atsushi421:
## PR Type

- New Feature

## Related Links

- [agnocast_cie_thread_configurator: spawn_non_ros2_thread](https://github.com/autowarefoundation/agnocast/blob/5d69d9e918231160b41a5708bc287738bfb8e892/src/agnocast_cie_thread_configurator/include/agnocast_cie_thread_configurator/cie_thread_configurator.hpp#L33)

## Description

Registers the non-ROS 2 threads on the hesai LiDAR packet path with `cie_thread_configurator`, so that their scheduling policy can be managed together with the CIE-managed callback threads (introduced in #418). Two threads are covered:

- **UDP receive thread** (`UdpSocket`): the thread that receives and dispatches LiDAR packets when running against a real sensor (`launch_hw:=true`). Registered as `nebula_hesai_udp_receiver@<frame_id>`.
- **Packets publish thread** (`SingleConsumerProcessor` used by `HesaiDecoderWrapper::packets_pub_thread_`): registered as `nebula_hesai_packets_pub@<frame_id>`.

The `frame_id` suffix keeps thread names unique per sensor, as required by `cie_thread_configurator`.

Design decisions:

- `nebula_core` and `nebula_hesai_hw_interfaces` stay free of Agnocast dependencies. `UdpSocket::subscribe()` and `SingleConsumerProcessor` gain an optional, middleware-agnostic `thread_factory` argument (default: a plain `std::thread` is created), and `HesaiHwInterface` gains an optional constructor parameter that carries the UDP thread factory down to `subscribe()`.
- Both Agnocast-aware factories are built only in the ROS wrapper layer (`nebula_hesai`), via the private helper `src/cie_thread_factory.hpp` (`make_cie_thread_factory`). This helper is the single `#ifdef USE_AGNOCAST_ENABLED` site in this PR and returns `nullptr` (callers create plain `std::thread`s) when Agnocast is disabled. `hw_interface_wrapper.cpp` injects the UDP factory (thread name built from `config->frame_id`) and `decoder_wrapper.cpp` injects the packets-publish factory. This avoids compiling the shared `SingleConsumerProcessor` template differently per package (a latent ODR hazard) and makes the factory path unit-testable without an Agnocast build. Other sensor vendors (velodyne, robosense, continental) and the hw_monitor diagnostics processor are unchanged and keep creating plain `std::thread`s.
- The new `agnocast_cie_thread_configurator` dependency is conditional (`$ENABLE_AGNOCAST == 1`) and lives only in `nebula_hesai`, linked `PRIVATE`: it is used only inside the wrapper sources and is not part of the package's exported interface.
- If a factory returns a non-joinable thread, `UdpSocket` / `SingleConsumerProcessor` throw instead of silently running without a receiver/consumer thread. Because the hw interface can be stopped and restarted, `hesai_hw_interface.cpp` copies (rather than moves) the stored factory into `subscribe()`; the factory is invoked again on each relaunch (the previous thread is joined first, so two factory-created threads never coexist).
- Everything is guarded by the existing `ENABLE_AGNOCAST` build-time convention: without the env var (or with a null `thread_factory`), behavior is byte-identical to before.
- If no configurator is listening, the announcement times out after 1 s with a warning and the thread runs normally (non-fatal by design in `send_non_ros_thread_info`).

Note: in rosbag replay mode (`launch_hw:=false`) neither thread is spawned; packet decoding runs inline in CIE-managed subscription callbacks, so this change only affects real-sensor operation.

## Review Procedure

1. Build with and without `ENABLE_AGNOCAST=1` and confirm both succeed. The factory paths are covered by unit tests in `test_udp.cpp` (`UdpSocket`) and `test_single_consumer_processor.cpp` (`SingleConsumerProcessor`), neither of which requires Agnocast.
2. With `ENABLE_AGNOCAST=1` builds, start the thread configurator (`ros2 run agnocast_cie_thread_configurator thread_configurator_node --prerun`), then launch a hesai node without sensor hardware:
   `launch_hw:=true udp_only:=true host_ip:=127.0.0.1` (udp_only skips the TCP connection, and the UDP socket binds locally).
   The configurator log should show both announcements:
   ```
   Received NonRosThreadInfo: tid=... | nebula_hesai_packets_pub@hesai
   Received NonRosThreadInfo: tid=... | nebula_hesai_udp_receiver@hesai
   ```
3. Regression: a full logging-simulator run (bag replay, `launch_hw:=false`) decodes normally (verified on pilot-auto.x2.4.4: decode rate matches the replay rate, no process deaths).

- [x] Assign PR to reviewer

## CI Checks

### DIFF
--- a/src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
+++ b/src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
@@ -271,7 +271,8 @@ public:
    * Has to be called on a bound socket (`bind()` has to have been called before).
    *
    * @param callback The function to be executed for each received packet.
-   * @param thread_factory Optional factory used to create the receiver thread.
+   * @param thread_factory Optional factory used to create the receiver thread. If not given, a
+   * plain `std::thread` is created.
    */
   UdpSocket & subscribe(callback_t && callback, thread_factory_t thread_factory = nullptr)
   {
@@ -392,6 +393,8 @@ private:
 
     if (thread_factory_) {
       receive_thread_ = thread_factory_(std::move(receive_loop));
+      // A factory returning a non-joinable thread would leave the socket silently dead.
+      assert(receive_thread_.joinable());
     } else {
       receive_thread_ = std::thread(std::move(receive_loop));
     }
--- a/src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
+++ b/src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
@@ -260,17 +260,24 @@ public:
 
   using callback_t = std::function<void(std::vector<uint8_t> & data, const RxMetadata & metadata)>;
 
+  /// Creates the receiver thread from a given thread body. Allows callers to integrate the
+  /// receiver thread with external thread management frameworks (e.g. to control its scheduling
+  /// policy). If not given, a plain `std::thread` is created.
+  using thread_factory_t = std::function<std::thread(std::function<void()> && thread_body)>;
+
   /**
    * @brief Register a callback for processing received packets and start the receiver thread. The
    * callback will be called for each received packet, and will be executed in the receive thread.
    * Has to be called on a bound socket (`bind()` has to have been called before).
    *
    * @param callback The function to be executed for each received packet.
+   * @param thread_factory Optional factory used to create the receiver thread.
    */
-  UdpSocket & subscribe(callback_t && callback)
+  UdpSocket & subscribe(callback_t && callback, thread_factory_t thread_factory = nullptr)
   {
     unsubscribe();
     callback_ = std::move(callback);
+    thread_factory_ = std::move(thread_factory);
     launch_receiver();
     return *this;
   }
@@ -316,7 +323,7 @@ public:
   UdpSocket(UdpSocket && other) noexcept
   : sock_fd_((other.unsubscribe(), std::move(other.sock_fd_))), config_(other.config_)
   {
-    if (other.callback_) subscribe(std::move(other.callback_));
+    if (other.callback_) subscribe(std::move(other.callback_), std::move(other.thread_factory_));
   };
 
   UdpSocket & operator=(const UdpSocket &) = delete;
@@ -328,7 +335,7 @@ private:
     assert(callback_);
 
     running_ = true;
-    receive_thread_ = std::thread([this]() {
+    std::function<void()> receive_loop = [this]() {
       std::vector<uint8_t> buffer;
       DropMonitor drop_monitor{};
       PerfCounters current_packet_perf_counters{};
@@ -381,7 +388,13 @@ private:
 
         callback_(buffer, metadata);
       }
-    });
+    };
+
+    if (thread_factory_) {
+      receive_thread_ = thread_factory_(std::move(receive_loop));
+    } else {
+      receive_thread_ = std::thread(std::move(receive_loop));
+    }
   }
 
   static void get_receive_metadata(msghdr & msg, RxMetadata & metadata, DropMonitor & drop_monitor)
@@ -421,6 +434,7 @@ private:
   std::atomic_bool running_{false};
   std::thread receive_thread_;
   callback_t callback_;
+  thread_factory_t thread_factory_;
 };
 
 }  // namespace nebula::drivers::connections
--- a/src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
+++ b/src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
@@ -21,6 +21,7 @@
 
 #include <nebula_core_common/util/errno.hpp>
 #include <nebula_core_common/util/expected.hpp>
+#include <nebula_core_common/util/thread_factory.hpp>
 #include <nebula_core_hw_interfaces/connections/socket_utils.hpp>
 
 #include <arpa/inet.h>
@@ -260,13 +261,11 @@ public:
 
   using callback_t = std::function<void(std::vector<uint8_t> & data, const RxMetadata & metadata)>;
 
-  /// Creates the receiver thread from a given thread body. Allows callers to integrate the
-  /// receiver thread with external thread management frameworks (e.g. to control its scheduling
-  /// policy). If not given, a plain `std::thread` is created.
+  /// Creates the receiver thread from a given thread body.
   /// Note: the factory is invoked again each time the receiver thread is relaunched, i.e. when a
   /// subscribed socket is re-subscribed or moved (the previous thread is joined first, so two
   /// factory-created threads never coexist).
-  using thread_factory_t = std::function<std::thread(std::function<void()> && thread_body)>;
+  using thread_factory_t = util::thread_factory_t;
 
   /**
    * @brief Register a callback for processing received packets and start the receiver thread. The
@@ -274,11 +273,13 @@ public:
    * Has to be called on a bound socket (`bind()` has to have been called before).
    *
    * @param callback The function to be executed for each received packet.
-   * @param thread_factory Optional factory used to create the receiver thread. If not given, a
-   * plain `std::thread` is created.
+   * @param thread_factory The factory used to create the receiver thread. Must not be null.
    */
-  UdpSocket & subscribe(callback_t && callback, thread_factory_t thread_factory = nullptr)
+  UdpSocket & subscribe(
+    callback_t && callback, thread_factory_t thread_factory = util::StdThreadFactory{})
   {
+    if (!thread_factory) throw UsageError("Thread factory must not be null");
+
     unsubscribe();
     callback_ = std::move(callback);
     thread_factory_ = std::move(thread_factory);
@@ -394,15 +395,11 @@ private:
       }
     };
 
-    if (thread_factory_) {
-      receive_thread_ = thread_factory_(std::move(receive_loop));
-      if (!receive_thread_.joinable()) {
-        // Roll back so the socket is left in a consistent unsubscribed state.
-        running_ = false;
-        throw UsageError("Thread factory must return a joinable thread");
-      }
-    } else {
-      receive_thread_ = std::thread(std::move(receive_loop));
+    receive_thread_ = thread_factory_(std::move(receive_loop));
+    if (!receive_thread_.joinable()) {
+      // Roll back so the socket is left in a consistent unsubscribed state.
+      running_ = false;
+      throw UsageError("Thread factory must return a joinable thread");
     }
   }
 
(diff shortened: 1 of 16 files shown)


src/nebula_core/nebula_core_ros/include/nebula_core_ros/single_consumer_processor.hpp
```diff
@@ -52,7 +60,13 @@ class SingleConsumerProcessor
...
     }
 
-    consumer_thread_ = std::thread(&SingleConsumerProcessor::consumer_loop, this);
+    if (thread_factory) {
+      consumer_thread_ = thread_factory([this]() { consumer_loop(); });
+      // A factory returning a non-joinable thread would leave the processor silently dead.
+      assert(consumer_thread_.joinable());
+    } else {
+      consumer_thread_ = std::thread(&SingleConsumerProcessor::consumer_loop, this);
+    }
```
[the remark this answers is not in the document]
atsushi421: Fixed in 4d82a24: replaced the assert with a runtime check that throws std::invalid_argument, consistent with the constructor's existing validations.

src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
```diff
@@ -381,7 +392,15 @@ class UdpSocket
...
-    });
+    };
+
+    if (thread_factory_) {
+      receive_thread_ = thread_factory_(std::move(receive_loop));
+      // A factory returning a non-joinable thread would leave the socket silently dead.
+      assert(receive_thread_.joinable());
+    } else {
+      receive_thread_ = std::thread(std::move(receive_loop));
+    }
```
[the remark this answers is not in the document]
atsushi421: Fixed in 4d82a24: the check is now a runtime one that resets running_ to false (so the socket is left in a consistent unsubscribed state) and throws UsageError.

mojomex (changes requested):
Thank you! Exciting stuff.

I agree with almost all of the implementation path:
* factory pattern is clean
* interface boundaries are clean

The only feedback I have is that I'd rather have the ROS wrappers (`nebula_hesai`) create all thread factories. Currently, it only does so for the SingleConsumerProcessor, but not for the HW interface/UDP thread.

Rationale: Agnocast is a middleware, and we want to keep lower packages middleware-independent, including their build scripts. That is what we did for logging as well, so having Agnocast align would be best!

atsushi421:
@mojomex 
Thank you for your review and suggestion!
I have addressed it in https://github.com/tier4/nebula/pull/460/commits/5d059c097a4a2f20d3e93db94f6c2fa1b9739039, so please review again.

src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
```diff
@@ -260,17 +260,28 @@ class UdpSocket
...
    * @brief Register a callback for processing received packets and start the receiver thread. The
    * callback will be called for each received packet, and will be executed in the receive thread.
    * Has to be called on a bound socket (`bind()` has to have been called before).
    *
    * @param callback The function to be executed for each received packet.
+   * @param thread_factory Optional factory used to create the receiver thread. If not given, a
+   * plain `std::thread` is created.
    */
-  UdpSocket & subscribe(callback_t && callback)
+  UdpSocket & subscribe(callback_t && callback, thread_factory_t thread_factory = nullptr)
```
mojomex: This is quite similar to the _bastard injection_ pattern. I'd prefer if we had a `StdThreadFactory` in `nebula_core_common` which callers can pass into here, and banning `nullptr` here. This makes the internals of the UDP code cleaner.

src/nebula_core/nebula_core_hw_interfaces/test/common/test_udp.cpp
```diff
@@ -215,6 +216,34 @@ TEST(TestUdp, TestMoveable)
...
+      n_factory_calls++;
+      return std::thread(std::move(thread_body));
+    });
+
+  ASSERT_EQ(n_factory_calls, 1);
+
+  auto err_no_opt = udp_send(g_localhost_ip, g_host_port, payload);
+  if (err_no_opt.has_value()) GTEST_SKIP() << strerror(err_no_opt.value());
+
+  std::this_thread::sleep_for(100ms);
```
mojomex: Fixed wait time either cause
* inflated runtimes when chosen big
* flaky tests if chosen small

Consider doing this via std::promise (set in receive callback) and std::future (wait_for with timeout) instead.

src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
```diff
@@ -260,17 +260,28 @@ class UdpSocket
 
   using callback_t = std::function<void(std::vector<uint8_t> & data, const RxMetadata & metadata)>;
 
+  /// Creates the receiver thread from a given thread body. Allows callers to integrate the
+  /// receiver thread with external thread management frameworks (e.g. to control its scheduling
+  /// policy). If not given, a plain `std::thread` is created.
+  /// Note: the factory is invoked again each time the receiver thread is relaunched, i.e. when a
+  /// subscribed socket is re-subscribed or moved (the previous thread is joined first, so two
+  /// factory-created threads never coexist).
+  using thread_factory_t = std::function<std::thread(std::function<void()> && thread_body)>;
```
mojomex: As it's being reused in other places, it should probably be absorbed into `nebula_core_common`

src/nebula_core/nebula_core_ros/include/nebula_core_ros/single_consumer_processor.hpp
```diff
@@ -38,10 +38,18 @@ class SingleConsumerProcessor
...
+  using thread_factory_t = std::function<std::thread(std::function<void()> && thread_body)>;
+
   /// @brief Constructor
   /// @param callback The callback function to execute on each item
   /// @param max_queue_size The maximum size of the queue
-  explicit SingleConsumerProcessor(callback_t callback, size_t max_queue_size)
+  /// @param thread_factory Optional factory used to create the consumer thread. If not given, a
+  /// plain `std::thread` is created.
+  explicit SingleConsumerProcessor(
+    callback_t callback, size_t max_queue_size, thread_factory_t thread_factory = nullptr)
```
mojomex: Same bastard injection comment as above.

src/nebula_core/nebula_core_ros/include/nebula_core_ros/cie_thread_factory.hpp
mojomex, line 1: Since this is not Hesai specific, please put it into `nebula_core_ros` instead.

src/nebula_hesai/nebula_hesai_hw_interfaces/src/hesai_hw_interface.cpp
```diff
@@ -239,10 +243,12 @@ Status HesaiHwInterface::sensor_interface_start()
 
   udp_socket_.emplace(std::move(builder).bind());
 
+  // Copy rather than move the factory: the interface can be stopped and restarted.
```
mojomex: The UdpSocket class stores the factory as a member, so it is persisted across unsubscribes/resubscribes and we can move here.

Let's add that test case to test_udp 🙏

mojomex (commented):
Thanks for addressing the feedback!
I have some more small comments, please have a look once more 🙇

src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
atsushi421: Fixed in 5ea325c: added StdThreadFactory (together with the thread_factory_t alias) to nebula_core_common/util/thread_factory.hpp. subscribe() now defaults to StdThreadFactory{} and throws UsageError when a null factory is passed, so the internals no longer branch on nullptr. I kept the default argument so callers that do not care about thread management (the other vendors' HW interfaces) stay unchanged; let me know if you would rather make the parameter required.

src/nebula_core/nebula_core_hw_interfaces/test/common/test_udp.cpp
atsushi421: Fixed in 5ea325c: the test now waits on a std::future (the promise is set in the receive callback) with wait_for(g_send_receive_timeout) instead of a fixed sleep.

src/nebula_core/nebula_core_hw_interfaces/include/nebula_core_hw_interfaces/connections/udp.hpp
atsushi421: Fixed in 5ea325c: thread_factory_t now lives in nebula_core_common/util/thread_factory.hpp together with StdThreadFactory, and both UdpSocket and SingleConsumerProcessor alias it from there.

src/nebula_core/nebula_core_ros/include/nebula_core_ros/single_consumer_processor.hpp
atsushi421: Fixed in 5ea325c: same treatment as UdpSocket. The constructor now defaults to StdThreadFactory{} and throws std::invalid_argument on a null factory, removing the internal branch.

src/nebula_core/nebula_core_ros/include/nebula_core_ros/cie_thread_factory.hpp
atsushi421: Fixed in 5ea325c: moved to nebula_core_ros/cie_thread_factory.hpp, following the same pattern as agnocast_wrapper/nebula_agnocast_wrapper.hpp (header-only behind USE_AGNOCAST_ENABLED, so nebula_core_ros's build scripts stay Agnocast-free; the compile definition and the PRIVATE link remain in nebula_hesai). With null factories now banned, it returns StdThreadFactory{} instead of nullptr when Agnocast support is disabled.

src/nebula_hesai/nebula_hesai_hw_interfaces/src/hesai_hw_interface.cpp
atsushi421: Added the persistence test in 5ea325c (TestThreadFactoryPersistsAcrossResubscribe): the stored factory is reused across unsubscribe, re-subscribe and move without the caller re-passing it. Regarding the move here, I kept the copy: sensor_interface_start() can be called again after sensor_interface_stop(), and each call re-creates the socket via udp_socket_.emplace(), destroying the previous socket together with its stored factory. The interface-level member is what keeps the factory available for those restarts, and since null factories are now rejected, moving it would make a second start throw. I clarified the code comment accordingly; happy to restructure if you would prefer the socket to be created only once instead.

src/nebula_hesai/nebula_hesai/CMakeLists.txt
```diff
@@ -46,6 +51,12 @@ ament_target_dependencies(
 if("$ENV{ENABLE_AGNOCAST}")
   message(STATUS "Agnocast support is enabled")
   ament_target_dependencies(hesai_ros_wrapper PUBLIC agnocastlib)
+  # Used only by nebula_core_ros/cie_thread_factory.hpp, whose Agnocast branch
+  # is compiled into this package's translation units; not part of the exported
+  # interface.
```
mojomex, line 56: Hm, ideally we would have agnocast_cie_thread_configurator as an `INTERFACE` dependency in `nebula_core_ros` but since this is the only user for now, I think it's fine.

mojomex (approved):
LGTM! Confirmed it operates correctly on PCAPs without Agnocast.

PR #464, opened 2026-07-22
fix(nebula_hesai): guard sensor setup with try/catch and retry

mojomex:
## PR Type

- Bug Fix

## Related Links

- Stacked on top of #463 (must merge first). This PR targets that branch, not `main`.

## Description

Follow-up to #463. Once the throwaway `std::thread`+`join()` wrappers were removed from `check_and_set_config`, a comms fault during sensor setup surfaces as a normal `std::exception` at the call site instead of calling `std::terminate()`. But the call site in `HesaiHwInterfaceWrapper` still did not handle it, so a transient timeout during startup (e.g. `get_lidar_range()` → `"PTC Error: Timeout;"`) would propagate out of node construction — and out of the `on_config_change` parameter callback at runtime.

This PR adds a `configure_sensor()` helper that:

- wraps `check_and_set_config()` in `try/catch (const std::exception &)` and, when `retry_hw` is set, retries up to `g_hw_config_max_attempts` (5) times with `g_hw_retry_delay` (8 s) backoff — the same cadence already used for the TCP-connect step (single best-effort attempt when `retry_hw` is false);
- **throws `std::runtime_error` if configuration still cannot be applied after all attempts**, rather than running a misconfigured sensor.

Failure handling differs by caller, by design:

- **Startup (constructor):** the exception propagates, so the node fails to construct. A sensor we cannot configure is treated as fatal rather than started misconfigured.
- **Runtime (`on_config_change`):** this runs inside an `add_on_set_parameters_callback` handler, where an unhandled exception would take down the *running* node. So it is caught and logged, and the node keeps running with the previously applied configuration.

Retry count and backoff are named constants (`g_hw_config_max_attempts`, `g_hw_retry_delay`); the backoff constant also replaces the previously duplicated `8000 ms` literal in the TCP-connect retry loop. `retry_hw` is stored as a member so both the constructor and `on_config_change` share the retry behaviour.

## Review Procedure

- Confirm a transient comms fault at startup is retried (when `retry_hw`) and, if it persists, aborts node construction with a clear error rather than starting a misconfigured sensor.
- Confirm a persistent failure during a runtime parameter change is logged and does **not** crash the running node.
- Confirm `retry_hw=false` keeps single-attempt behaviour.
- Confirm the success path is unchanged.

## Remarks

Part of a stack hardening the Hesai PTC paths against sensor/network faults. A separate follow-up will bound the response-length allocation in `send_receive` and guard the HW-monitor's `SingleConsumerProcessor` callback.

## CI Checks

---
Built locally with `colcon build --packages-up-to nebula_hesai` ✅. Formatted with pinned clang-format v22.1.4 via pre-commit.

### DIFF
--- a/src/nebula_hesai/nebula_hesai/include/nebula_hesai/hw_interface_wrapper.hpp
+++ b/src/nebula_hesai/nebula_hesai/include/nebula_hesai/hw_interface_wrapper.hpp
@@ -42,6 +42,14 @@ public:
   [[nodiscard]] std::shared_ptr<const HesaiInventoryBase> inventory() const;
 
 private:
+  /// @brief Apply the sensor configuration via the HW interface, retrying on transient comms
+  /// faults. Retries are bounded by g_hw_config_max_attempts when retry_hw is enabled (a single
+  /// attempt otherwise).
+  /// @throws std::runtime_error if the configuration cannot be applied after all attempts. Callers
+  /// decide whether that is fatal: the constructor lets it propagate (aborting startup), while
+  /// on_config_change catches it to keep a running node alive.
+  void configure_sensor();
+
   std::shared_ptr<drivers::HesaiHwInterface> hw_interface_;
   std::shared_ptr<const HesaiInventoryBase> inventory_;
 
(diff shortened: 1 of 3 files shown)


shtokuda (approved):
LGTM

PR #469, opened 2026-07-23
feat(nebula_core_ros): make SyncToolingWorker compatible with agnocast_wrapper::Node

Koichi98:
## PR Type

- Improvement

## Related Links

- Stacked base for the hesai and continental driver migrations (reviewed in Koichi98/nebula#1 and Koichi98/nebula#2, which merge into this branch). See the atomic-merge note below.

## Description

Make the shared `SyncToolingWorker` compatible with `agnocast_wrapper::Node`.

**Why hard-code `agnocast_wrapper::Node *` instead of a `NodeT` template:** `SyncToolingWorker` holds a publisher *member* whose type depends on the backend. Unlike `RateBoundStatus` / `LivenessMonitor`, which take the node only in a templated constructor and keep no node-typed member, a templated constructor can't type that member — the only template option would be making the whole class a template with a `decltype(node->create_publisher<...>(...))`-deduced member. That would only add plain `rclcpp::Node` support, which no caller needs: `SyncToolingWorker` is used only by `HesaiRosWrapper` and `ContinentalARS548RosWrapper`, and both migrate to `agnocast_wrapper::Node`. Since the wrapper Node already abstracts the `ENABLE_AGNOCAST=0/1` backend, `agnocast_wrapper::Node *` + `NEBULA_PUBLISHER_PTR` is simpler and sufficient.

**⚠️ Must merge atomically with the driver migrations:** without the templated constructor, `SyncToolingWorker` no longer accepts a plain `rclcpp::Node *`. On `main` the drivers are still `rclcpp::Node` and pass one, so merging this branch to `main` alone breaks the build; the migrations pass an `agnocast_wrapper::Node *` that the old constructor rejects. The hesai and continental migrations therefore merge into this branch first, and this branch goes to `main` only once both are integrated.

## Review Procedure

Build with `ENABLE_AGNOCAST=0` and `=1` after the hesai / continental migrations are merged into this branch.

## Remarks

None.

- [ ] Assign PR to reviewer

## CI Checks


mojomex (approved):
LGTM. Confirmed Nebula compiles&runs with and without Agnocast enabled.

Tested with rosbags for both OT128 and ARS548.

PR #493, opened 2026-09-15
fix(nebula_hesai_decoders): drop scans whose timestamp was never set

ogrex:
## Problem

`DecodeFrame::scan_timestamp_ns` is initialised to 0 and is written only by the scan cutter's set-timestamp callback. Two things have to line up for that callback never to fire:

1. `ScanCutter::initialize_state()` issues it only when the first decoded block intersects the configured FoV — if the decoder starts up with the sensor azimuth outside the FoV, neither branch sets a timestamp.
2. `FsmCutInFov` — selected whenever `cut_angle` lies strictly inside the FoV — resets a buffer's timestamp only on a state transition, and always on the buffer it is *entering*:

```cpp
case State::F0:
  switch (state_after) {
    case State::F1:
      actions.reset_timestamp_buffer = 1;   // entering
      actions.emit_scan_buffer = 0;         // leaving — timestamp still unset
```

So the buffer that was already active at initialisation never receives a timestamp, and is emitted as-is at the first cut. `FsmCutAtFovEnd` does not have this hole: there the buffer being emitted always had its timestamp set on FoV entry.

The same applies after `reset_decoder()` rebuilds the decoder on a non-monotonic packet during replay.

Such a scan is doubly wrong: it is stamped 0, and having started mid-revolution it covers only the arc from FoV entry to the cut angle. `publish_cloud()` does not catch it — it tests only `stamp.sec < 0`.

## Impact

Observed on a multi-lidar replay, on the sensors whose cut angle lies inside their FoV. The emitted clouds spanned exactly the arc that model predicts, were confined to that sector, and carried a few percent of the neighbouring clouds' point counts.

The zero then propagates: the concatenation node takes the minimum input stamp, so a single such scan zeroes the whole concatenated cloud, and detectors copy that header into their output. Downstream this is enough to poison consumers that derive latency from the header — in our case a tracker whose unbounded latency statistics absorbed `now - 0`, stalling for tens of seconds.

## Fix

Drop the frame rather than publish a timestamp known to be wrong, clearing the buffer as the normal path does so the next scan starts clean.

Resetting the blockage mask needs a plugin entry point, since `BlockageMask::reset()` is private to `BlockageMaskPlugin`; this adds one that resets without invoking the callback.

## Tests

`colcon build --packages-select nebula_hesai_decoders nebula_hesai` passes. The behaviour change is confined to scans that would otherwise be published with a timestamp of 0, which are invalid by construction.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
```diff
@@ -281,6 +281,22 @@ class HesaiDecoder : public HesaiScanDecoder
...
 
     auto & completed_frame = frame_buffers_[buffer_index];
+
+    // `scan_timestamp_ns` is initialised to 0 and is only written by the scan cutter's
+    // set-timestamp callback, which is not issued when the decoder initialises (or is reset by
+    // `reset_decoder()` on a non-monotonic packet) while the sensor azimuth lies outside the
+    // configured FoV. The buffer is nevertheless emitted at the next cut, so a scan can reach
+    // subscribers stamped 0 -- and being a partial revolution, it is short as well. Downstream
+    // nodes copy the header verbatim, so the zero propagates into the perception pipeline.
+    // Drop such a scan rather than publishing a timestamp that is known to be wrong.
```
mojomex: Please make this description more compact - 2-3 lines at most.

mojomex (commented):
Thank you! Code is almost good to go.
Could you also add a unit test that guarantees the behavior in your described edge case is now correct?

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
ogrex: addressed in [3213e19](https://github.com/tier4/nebula/pull/493/commits/3213e199b2d25c70c004d6c6ba3ea431f0e1eb99)

src/nebula_hesai/nebula_hesai_decoders/include/nebula_hesai_decoders/decoders/hesai_decoder.hpp
```diff
@@ -281,6 +281,17 @@ class HesaiDecoder : public HesaiScanDecoder
     did_scan_complete_ = true;
 
     auto & completed_frame = frame_buffers_[buffer_index];
+
+    // A scan initialised outside the FoV never gets its timestamp set, yet is still emitted at
+    // the next cut. Drop it: the stamp is known to be wrong and downstream nodes copy it verbatim.
```
mojomex, line 286:
```suggestion
    // the next cut. Drop it as its timestamp is meaningless to downstream consumers.
```

mojomex (approved):

PR #494, opened 2026-09-17
fix(nebula_core_ros): build the synced agnocast_wrapper Node implementation

Koichi98:
## PR Type

- Bug Fix

## Related Links

- Introduced by https://github.com/tier4/nebula/pull/469, worked around downstream by https://github.com/tier4/pilot-auto.x2/pull/5596
- Follow-up: https://github.com/tier4/nebula/pull/495

## Description

`sync-files.yaml` mirrors the `agnocast_wrapper` headers from `autoware.core` but not the sources, so `Node`'s non-template members — declared in `node.hpp`, defined in `src/node.cpp` — were defined nowhere. An `ENABLE_AGNOCAST=1` build links fine and fails only at component load time:

```
Failed to load library: libhesai_ros_wrapper.so: undefined symbol:
_ZNK6nebula16agnocast_wrapper4Node23get_node_base_interfaceEv
```

Sync `src/node.cpp` and compile it into `nebula_core_ros`. `USE_AGNOCAST_ENABLED` goes `PUBLIC` with it: it picks which `Node` class this library compiles into itself, and `nebula_continental` was compiling the other one. Works with the agnocastlib 2.3.5 we pin today.

## Review Procedure

Build with `ENABLE_AGNOCAST=1` and check the resolved link (undefined entries legitimately remain in `nm` output):

```bash
ldd -r install/nebula_hesai/lib/libhesai_ros_wrapper.so | grep "undefined symbol"   # expect no output
```

## Remarks

Verified on an `ENABLE_AGNOCAST=1` build of pilot-auto.x2: the 30 missing `Node::` definitions are exported, hesai and continental resolve fully, and the six ARS548 radar nodes stay up in the logging simulator. Note that nebula CI never sets `ENABLE_AGNOCAST`, so this code path is not built by CI.

- [ ] Assign PR to reviewer

## CI Checks


mojomex (approved):
LGTM!
Smoke tested with/without Agnocast in a fresh Docker.

PR #495, opened 2026-09-18
chore(nebula_core_ros): follow upstream's agnocast_wrapper layout

Koichi98:
## PR Type

- Improvement

## Related Links

- Upstream package: `autowarefoundation/autoware.core` `common/autoware_agnocast_wrapper`
- Builds on https://github.com/tier4/nebula/pull/494 (merge that first)
- Needs agnocastlib >= 2.4.0, so it lands together with the agnocast bump on the consuming side

## Description

`sync-files.yaml` mirrors five headers of `autoware::agnocast_wrapper`, a layout upstream has since left behind: the monolithic header was split into per-topic headers, `use_agnocast()` moved into `src/runtime.cpp`, and the register_node machinery was added. Syncing those five files alone would import an aggregator header that includes files we do not have.

This PR syncs every file of the package instead, headers and sources alike, plus the register_node macro, its config extras and its template. The one exception is `polling_subscriber.hpp`, the `autoware_utils_rclcpp` bridge — nothing else in the package includes it, so nebula still never pulls `autoware_utils_rclcpp`.

`runtime.cpp` joins `node.cpp` in `nebula_core_ros` and is always compiled, since it defines `use_agnocast()` in both builds.

The Hesai and Continental nodes now register through `nebula_agnocast_wrapper_register_node()`. `agnocast_components_register_node()` with an AgnocastOnly executor generates a main that calls `agnocast::init()` alone, which brings up no rclcpp context, so an `ENABLE_AGNOCAST=1` build aborted when run with `ENABLE_AGNOCAST=0`:

```
what(): failed to create guard condition: context argument is null
```

The synced macro's template picks the context at runtime instead. `nebula_core_ros` installs the macro and hands it to consumers through `CONFIG_EXTRAS`, so no package outside nebula is needed. glog comes along for the failure signal handler the generated executable installs; like agnocastlib it is only pulled in when `ENABLE_AGNOCAST` is set.

## Review Procedure

Requires **agnocastlib >= 2.4.0** — `agnocast::GenericPublisher`, `GenericSubscription` and `agnocast_service_event_publisher.hpp` do not exist in 2.3.5. `package.xml` declares that bound, as upstream does.

Build with `ENABLE_AGNOCAST=1`, then start one driver twice to see the runtime switch:

```bash
ENABLE_AGNOCAST=1 ros2 run nebula_hesai hesai_ros_wrapper_node   # "Starting node with agnocast mode"
ENABLE_AGNOCAST=0 ros2 run nebula_hesai hesai_ros_wrapper_node   # "Starting node with rclcpp mode"
```

An `ENABLE_AGNOCAST=0` build is unaffected: the macro falls back to `rclcpp_components_register_node`, and none of the agnocast branches of the headers are compiled.

## Remarks

Verified on a full `ENABLE_AGNOCAST=1` build of pilot-auto.x2 (agnocastlib 2.4.0, ROS 2 Jazzy). Both drivers start in either mode from the same binary, and the logging simulator runs with no process deaths in both: 103 components loaded under `ENABLE_AGNOCAST=1` and 82 under `ENABLE_AGNOCAST=0`, with the six ARS548 radar nodes up in both. Every synced file was also compiled standalone with no agnocast headers present, covering the `ENABLE_AGNOCAST=0` build.

Note that nebula CI never sets `ENABLE_AGNOCAST`, so this code path is not built by CI.

- [ ] Assign PR to reviewer

## CI Checks


mojomex (approved):
Please update Agnocast in `build_depends*.repos` to `2.4.0` or later.

Otherwise LGTM! Smoke tested with/without Agnocast.

Koichi98:
@mojomex 

> Please update Agnocast in build_depends*.repos to 2.4.0 or later.

Fixed in https://github.com/tier4/nebula/pull/495/commits/5450b183999a4259b8c5fadefe82dd79da81551d!

